PR #8, opened 2019-05-13
No need to run golangci ourselves, use the service instead.

geeknoid:


geeknoid:
Never mind, the service isn't quite up to snuff.

PR #9, opened 2019-05-13
Fix "make format" and "go generate"

geeknoid:
Fixes <https://github.com/istio/istio/issues/14054>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,3 +3,6 @@ lint:
 
 format:
 	@scripts/fmt.sh
+
+gen:
+	@go generate ./...
--- a/ctrlz/topics/assets/assets.go
+++ b/ctrlz/topics/assets/assets.go
@@ -12,6 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:generate $GOPATH/src/istio.io/istio/bin/go-bindata.sh --nocompress --nometadata --pkg assets -o assets.gen.go ./templates/...
+//go:generate ../../../scripts/go-bindata.sh --nocompress --nometadata --pkg assets -o assets.gen.go ./templates/...
 
 package assets
--- a/ctrlz/assets/assets.go
+++ b/ctrlz/assets/assets.go
@@ -12,6 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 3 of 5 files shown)


geeknoid:
This makes the fmt.sh and linters.sh repo-neutral. Once this is in, I'll update the other repos where I copied those files.

ctrlz/topics/assets/assets.go
```diff
@@ -12,6 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:generate $GOPATH/src/istio.io/istio/bin/go-bindata.sh --nocompress --nometadata --pkg assets -o assets.gen.go ./templates/...
+//go:generate ../../../scripts/go-bindata.sh --nocompress --nometadata --pkg assets -o assets.gen.go ./templates/...
```
ozevren: What is the root of this? Is it the invocation current-working directory, or is it he location of assets.go? (It looks like the latter).

Makefile
```diff
@@ -3,3 +3,6 @@ lint:
 
 format:
 	@scripts/fmt.sh
```
kyessenov: Maybe switch to `golangci-lint run --fix`?

scripts/linters.sh
```diff
@@ -16,11 +16,10 @@
```
kyessenov: I'm not sure about the value of this script. It could be incorporated straight into the makefile since it's just 3-4 shell commands.

ctrlz/topics/assets/assets.go
geeknoid: The current directory when this thing runs is set to the location of the .go file. So I can do "go generate ./..." from the root of the repo, or I can do "go generate" from the specific subdirectory and it works.

scripts/linters.sh
geeknoid: I expect this to be expanded over time. We need to call our custom linters here in particular. Before doing so, we need to continue doing a bit of code shuffling. I'd like to move linters to the tools repo and build them into docker images that we can use from other repos.

kyessenov: OK. It's just another line of `docker run` if the plan is to package all our linters into a docker container.

Makefile
geeknoid: I added "make fixlint" for this. Good idea.

PR #15, opened 2019-05-16
missing export of a constructor

kyessenov:
Signed-off-by: Kuat Yessenov <kuat@google.com>

### DIFF
--- a/attribute/stringmap.go
+++ b/attribute/stringmap.go
@@ -30,10 +30,11 @@ type StringMap struct {
 }
 
 // NewStringMap instantiates a new string map.
-func NewStringMap(name string) StringMap {
+func NewStringMap(name string, entries map[string]string, owner Bag) StringMap {
 	return StringMap{
 		name:    name,
-		entries: make(map[string]string, 1),
+		entries: entries,
+		owner:   owner,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


attribute/stringmap.go
```diff
@@ -30,10 +30,11 @@ type StringMap struct {
 }
 
 // NewStringMap instantiates a new string map.
-func NewStringMap(name string) StringMap {
+func NewStringMap(name string, entries map[string]string, owner Bag) StringMap {
```
ozevren: This seems like a dangerous constructor as entries will be shared with the caller. I don't see the usage of copyValue, but it looks like it is a private method and it seems ok to keep the inline construction model in the original code (as opposed to exposing a constructor that can cause data-sharing issues).

kyessenov: owner field is completely private. We don't have any other way to create a stringmap with an owner.

kyessenov: I can add a comment saying that the stringmap should take ownership of the provided map.

ozevren: how about just
```go
func NewMap(name string, owner Bag) StringMap {
}
```
then?

kyessenov: This is not supposed to be a mutable map. How do we set both owner and entries in proto bag (which is not in this package)?

kyessenov: Also, note that we already have `WrapStringMap` that accepts a map.

PR #18, opened 2019-05-22
feat(metrics docs): add collateral generation for metrics documentation

douglas-reid:
This PR adds the ability to export metrics documentation in HTML to the `collateral` package. Currently, only an OpenCensus-based export is provided. Output is an HTML snippet file (`metrics.html`) that consists of a `<table>` that includes name, type, and description. 

Looking for feedback on command name, file location, HTML structure.

Example screenshot of a rendered table for `mixs`:

image

### DIFF
--- a/collateral/control.go
+++ b/collateral/control.go
@@ -43,6 +43,7 @@ import (
 	"gopkg.in/yaml.v2"
 
 	"istio.io/pkg/annotations"
+	"istio.io/pkg/collateral/metrics"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/cobra/doc"
@@ -75,6 +76,10 @@ type Control struct {
 	// EmitHTMLFragmentWithFrontMatter controls whether to produce HTML fragments with Jekyll/Hugo front matter.
 	EmitHTMLFragmentWithFrontMatter bool
 
+	// EmitMetricsHTMLFragment controls whether to produce HTML fragments describing the metrics exported by the
+	// process.
+	EmitMetricsHTMLFragment bool
+
 	// ManPageInfo provides extra information necessary when emitting man pages.
 	ManPageInfo doc.GenManHeader
 }
@@ -98,6 +103,12 @@ func EmitCollateral(root *cobra.Command, c *Control) error {
 	if c.EmitHTMLFragmentWithFrontMatter {
 		if err := genHTMLFragment(root, c.OutputDir+"/"+root.Name()+".html"); err != nil {
 			return fmt.Errorf("unable to output HTML fragment file: %v", err)
+		}
+	}
+
+	if c.EmitMetricsHTMLFragment {
+		if err := genMetricsHTML(root, c.OutputDir+"/"+root.Name()+"/metrics.html"); err != nil {
+			return fmt.Errorf("unable to output metrics HTML fragment file: %v", err)
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


geeknoid:
This is great.

One issue though. Why did you go the route of outputting this as a separate file, as opposed to merely a section in the existing per-component file? If you look in control.go's genHTMLFragment function, it calls out to genVars and genAnnotations. This ends up producing a single per-command HTML page which can be rendered in the web site. So you see the page for "Pilot" and it lists all of its commands, its environment vars, its annotations, and its metrics.

So would you mind rejiggering the integration to follow this pattern?

douglas-reid:
@geeknoid I was of two minds here. I was thinking that it would be nice to have a central page that lists all of the metrics across components for Istio. It seemed like exporting to a separate file would make that easier.  Thoughts?

geeknoid:
Producing a centralized page is a good idea, but so is producing per-component stuff. So let's do both.

Output everything in a single file so we get the consistent component-level page. And then I'll a bit of voodoo in Hugo to automatically extract the metric info from those pages and auto-create a centralized page. I did that kind of stuff to produce adapter and template lists, so it's a workable pattern.

collateral/control.go
```diff
@@ -344,6 +355,20 @@ func genHTMLFragment(cmd *cobra.Command, path string) error {
 	return err
 }
 
+func genMetricsHTML(cmd *cobra.Command, path string) error {
```
golangcibot: `genMetricsHTML` - `cmd` is unused (from `unparam`)

collateral/metrics/opencensus_test.go
```diff
@@ -0,0 +1,121 @@
...
+	expected = `<h2 id=\"metrics\">Exported Metrics</h2>
+<table class=\"metrics\">
+<thead>
+<tr><th>Name</th><th>Type</th><th>Description</th></tr>
+</thead>
+<tbody>
+<tr><td>mixer_config_adapter_info_configs_total</td><td>LastValue</td><td>The number of known adapters in the current config.</td></tr>
+<tr><td>mixer_config_attributes_total</td><td>LastValue</td><td>The number of known attributes in the current config.</td></tr>
+<tr><td>mixer_config_handler_configs_total</td><td>LastValue</td><td>The number of known handlers in the current config.</td></tr>
+<tr><td>mixer_config_instance_config_errors_total</td><td>LastValue</td><td>The number of errors encountered during processing of the instance configuration.</td></tr>
```
golangcibot: line is 167 characters (from `lll`)

geeknoid: Maybe you should check in the test file as a .golden file and read it in from the rest, to avoid the line length issue.

collateral/control.go
```diff
@@ -732,3 +734,22 @@ func (g *generator) genAnnotations(root *cobra.Command) {
 	g.emit("</tbody>")
 	g.emit("</table>")
 }
+
+func (g *generator) genMetrics(root *cobra.Command) {
```
golangcibot: `(*generator).genMetrics` - `root` is unused (from `unparam`)

douglas-reid:
@geeknoid PTAL. Switched to having it be part of the `--html_fragment_with_front_matter=true` export.

douglas-reid:
@geeknoid how can we proceed here?  I didn't touch `filewatcher.go`, but that seems to be impacting the codecov calculations?

PR #20, opened 2019-05-28
Improve pkg/log

geeknoid:
- Add support for the "all" scope override. This lets you control the
output level of all scopes at once, instead of having to enumerate every
scope on the command-line.

- Eliminate problem with non-deterministic test coverage numbers. This happened
because the tests are modifying some global state. This global state is now
explicitly reset by the tests on startup to ensure a stable test execution.

- Improve how the Fatal* family of functions is tested so that code coverage of the
package can be boosted to 100%.

- Also, update common files.


log/config.go
```diff
@@ -194,65 +194,74 @@ func formatDate(t time.Time, enc zapcore.PrimitiveArrayEncoder) {
...
+		return err
+	}
+
+	// update the caller location setting of all listed scopes
+	sc := strings.Split(options.logCallers, ",")
+	overridePresent := false
+	for _, s := range sc {
+		if s == "" {
+			continue
+		} else if s == OverrideScopeName {
```
ozevren: Nit: you don't need else continuation. Might be simpler to have to two if blocks back to back.

log/config.go
```diff
@@ -270,10 +279,22 @@ func Configure(options *Options) error {
 		return err
 	}
 
-	if err = updateScopes(options, core, errSink); err != nil {
+	if err = updateScopes(options); err != nil {
```
ozevren: Would it be simpler to follow the patchTable model for the global method overrides?
There are few public entry points. They can call private counter-parts with a patch table, similar to how Mixer server code works. WDYT?

log/config.go
geeknoid: Done

geeknoid:
PTAL. Added a patch table. It's not as nice as in Mixer due to the fact this is using (by design) some global state. But it does encapsulate the replaceable bits.

PR #23, opened 2019-06-11
Fix bad HTML output.

geeknoid:
Extra backslashes are resulting in bad HTML.

Also, for Istio docs we only capitalize the first letter in headers, not every word.


geeknoid:
Pushing this in pronto, we need this in the release-1.2 branch to get good docs for the 1.2 release

PR #27, opened 2019-06-18
Explain how to see JSON output and suppress control plane check

esnible:
Resolves https://github.com/istio/istio/issues/14814

### DIFF
--- a/version/cobra.go
+++ b/version/cobra.go
@@ -109,7 +109,7 @@ func CobraCommandWithOptions(options CobraOptions) *cobra.Command {
 		},
 	}
 
-	cmd.Flags().BoolVarP(&short, "short", "s", false, "Use --short=false to generate JSON")
+	cmd.Flags().BoolVarP(&short, "short", "s", false, "Use --short=false to generate full version information")
 	cmd.Flags().StringVarP(&output, "output", "o", "", "One of 'yaml' or 'json'.")
 	if options.GetRemoteVersion != nil {
 		cmd.Flags().BoolVar(&remote, "remote", false, "Use --remote=false to suppress control plane check")
--- a/version/cobra.go
+++ b/version/cobra.go
@@ -109,10 +109,10 @@ func CobraCommandWithOptions(options CobraOptions) *cobra.Command {
 		},
 	}
 
-	cmd.Flags().BoolVarP(&short, "short", "s", false, "Displays a short form of the version information")
+	cmd.Flags().BoolVarP(&short, "short", "s", false, "Use --short=false to generate JSON")
 	cmd.Flags().StringVarP(&output, "output", "o", "", "One of 'yaml' or 'json'.")
 	if options.GetRemoteVersion != nil {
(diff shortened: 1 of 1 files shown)


version/cobra.go
```diff
@@ -109,10 +109,10 @@ func CobraCommandWithOptions(options CobraOptions) *cobra.Command {
 		},
 	}
 
-	cmd.Flags().BoolVarP(&short, "short", "s", false, "Displays a short form of the version information")
+	cmd.Flags().BoolVarP(&short, "short", "s", false, "Use --short=false to generate JSON")
```
howardjohn: `--short=false` is not Json, `-o json` is. I am not sure what to call it though..

geeknoid: How about calling the flag json?

howardjohn: But it isn't json. It will be something like:
```
version.BuildInfo{Version:"661ca4bcba7cf82208864d3f36a80e1d936712a5-dirty", GitRevision:"661ca4bcba7cf82208864d3f36a80e1d936712a5-dirty", User:"howardjohn", Host:"howardjohn.svl.corp.google.com", GolangVersion:"go1.12.5", DockerHub:"docker.io/istio", BuildStatus:"Modified", GitTag:"1.2.0-rc.0-109-g661ca4bcb
```

The json output is
```
{
  "clientVersion": {
    "version": "661ca4bcba7cf82208864d3f36a80e1d936712a5-dirty",
    "revision": "661ca4bcba7cf82208864d3f36a80e1d936712a5-dirty",
    "user": "howardjohn",
    "host": "howardjohn.svl.corp.google.com",
    "golang_version": "go1.12.5",
    "hub": "docker.io/istio",
    "status": "Modified",
    "tag": "1.2.0-rc.0-109-g661ca4bcb"
  }
}
```

ozevren: Seems like Go value output. Given that, the command-line message seems apt.

How do we use it? Rather, which output format do we use?

esnible: The original messages being replaced were created when the defaults were `false`.  Now that the defaults are `true`, users need to know to use `=false` to get the old behavior.

I had forgotten the output was Go source code format!

Instead of documenting the possibility of `--short=false` how about I add a new flag `--json` that actually outputs in JSON and that I deprecate `--short`?  Or we can leave it in, undeprecated.  I don't think anyone appreciates the Golang output and we should consider dumping it.

howardjohn: FWIW kubernetes uses the go source format. The go format is about 10x more compact than our json output, because we do pretty printing of the json so each remote component takes 10 lines compared to 1. We already have an `-o json` flag so I am not sure if we should add a `--json` flag as well, or if it is worth the deprecation effort.

If we changed short to output compact json instead of go source that seems fine to me though? I doubt anyone is depending on the format of it

esnible:
Why is `codecov/project` still waiting?  Is there a way to motivate the test to try again?

geeknoid:
The code coverage stuff is flaky.

I'll force merge...

PR #30, opened 2019-06-26
For istioctl version --short, coalese control plane version output

esnible:
Resolves https://github.com/istio/istio/issues/14794

For `istioctl version`, if all the control plane pods are running the same Istio version, the output now shows a single version line for the control plane instead of repeating that information.

This only applies to the short output, if the user wants the Go struct or JSON output the output is unchanged.

@geeknoid I was surprised `make lint` didn't work in this repo.  I ran `golangci-lint run` in the _version_ directory instead.


esnible:
The goal of this item is to allow `istioctl version` to output

```
client version: 1.2.0
control plane version: 1.2.0
```

... instead of showing the version of every pod in the control plane.

esnible:
@geeknoid Any idea why code coverage has not been reported?

PR #31, opened 2019-07-02
feat(monitoring pkg): add new common monitoring package

douglas-reid:
This PR adds a new common package for monitoring Istio components. It wraps the OpenCensus libraries with a developer-friendly API, allowing easy collateral generation and a consistent interface across components. There have been several PRs recently requesting such a monitoring package for broad `istio.io/istio` usage. This PR serves as a small refinement on the existing `istio.io/istio/pilot/pkg/monitoring` package, which it is intended to replace.

Examples have been provided to demonstrate the API in use.

### DIFF
--- a/monitoring/example_distribution_test.go
+++ b/monitoring/example_distribution_test.go
@@ -1,0 +1,36 @@
+// Copyright 2019 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package monitoring_test
+
+import "istio.io/pkg/monitoring"
+
+var (
+	method = monitoring.MustCreateLabel("method")
+
+	received_bytes = monitoring.NewDistribution(
+		"received_bytes_total",
+		"Distribution of received bytes by method",
+		[]float64{10, 100, 1000, 10000},
+		monitoring.WithLabels(method),
+		monitoring.WithUnit(monitoring.Bytes),
+	)
+)
+
+func init() {
+	monitoring.MustRegister(received_bytes)
+}
+
+func ExampleNewDistribution() {
+	received_bytes.With(method.Value("/projects/1")).Record(458)
+}
--- a/monitoring/example_distribution_test.go
+++ b/monitoring/example_distribution_test.go
@@ -18,7 +18,7 @@ import "istio.io/pkg/monitoring"
 var (
 	method = monitoring.MustCreateLabel("method")
 
-	received_bytes = monitoring.NewDistribution(
+	receivedBytes = monitoring.NewDistribution(
 		"received_bytes_total",
 		"Distribution of received bytes by method",
 		[]float64{10, 100, 1000, 10000},
@@ -28,9 +28,9 @@ var (
 )
 
 func init() {
-	monitoring.MustRegister(received_bytes)
+	monitoring.MustRegister(receivedBytes)
 }
 
 func ExampleNewDistribution() {
-	received_bytes.With(method.Value("/projects/1")).Record(458)
+	receivedBytes.With(method.Value("/projects/1")).Record(458)
 }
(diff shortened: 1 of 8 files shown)


monitoring/example_distribution_test.go
```diff
@@ -0,0 +1,36 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package monitoring_test
+
+import "istio.io/pkg/monitoring"
+
+var (
+	method = monitoring.MustCreateLabel("method")
+
+	received_bytes = monitoring.NewDistribution(
```
golangcibot: don't use underscores in Go names; var `received_bytes` should be `receivedBytes` (from `golint`)

monitoring/monitoring_test.go
```diff
@@ -0,0 +1,204 @@
...
+func (r registerFail) Register() error {
+	return errors.New("fail")
+}
+
+type testExporter struct {
+	rows map[string][]*view.Row
+}
+
+func (t *testExporter) ExportView(d *view.Data) {
+	for _, r := range d.Rows {
```
golangcibot: S1011: should replace loop with `t.rows[d.View.Name] = append(t.rows[d.View.Name], d.Rows...)` (from `gosimple`)

pitlv2109:
@douglas-reid Would you mind rebasing and getting this in, since a couple of PRs depend on this change?

douglas-reid:
Rebased. Thanks!

pitlv2109:
Sorry for the late question. Do `example_<metric_type>_test.go` files do anything? Changing the values of those files doesn't seem to affect the test outcome.

pitlv2109:
Replying here, since there is no Issues forum for this repo. I'm having some problem using this with metrics with tags. I would see a `Failed to export to Prometheus: "" is not a valid label name for metric <metric_name>`, whereas I didn't have any problem with a similar implementation of this OpenCensus abstraction in pkg/pilot or https://github.com/istio/istio/blob/345f0d4502f8685391dc3043cf208d7ec276ceae/security/pkg/monitoring/monitoring.go. I noticed that the data was there (by accessing it via the Row data structure), but for some reason, it couldn't be exported to Promethesus endpoint.

douglas-reid:
> Sorry for the late question. Do `example_<metric_type>_test.go` files do anything? Changing the values of those files doesn't seem to affect the test outcome.

The example files aren't actually doing any testing. Having them produce output just so it could be checked didn't seem all that useful to me. I wanted them more for simple documentation purposes. We can go back and add tests if you think that is needed. See also: https://blog.golang.org/examples

douglas-reid:
> Replying here, since there is no Issues forum for this repo. I'm having some problem using this with metrics with tags. I would see a `Failed to export to Prometheus: "" is not a valid label name for metric <metric_name>`, whereas I didn't have any problem with a similar implementation of this OpenCensus abstraction in pkg/pilot or https://github.com/istio/istio/blob/345f0d4502f8685391dc3043cf208d7ec276ceae/security/pkg/monitoring/monitoring.go. I noticed that the data was there (by accessing it via the Row data structure, so the test from this PR may not detect it?), but for some reason, it couldn't be exported to Promethesus endpoint.
> 
> @douglas-reid

Without more information, I'm not sure how to respond. You definitely can't have a label name be the empty string. That shouldn't be possible in the other abstraction either.

pitlv2109:
> > Replying here, since there is no Issues forum for this repo. I'm having some problem using this with metrics with tags. I would see a `Failed to export to Prometheus: "" is not a valid label name for metric <metric_name>`, whereas I didn't have any problem with a similar implementation of this OpenCensus abstraction in pkg/pilot or https://github.com/istio/istio/blob/345f0d4502f8685391dc3043cf208d7ec276ceae/security/pkg/monitoring/monitoring.go. I noticed that the data was there (by accessing it via the Row data structure, so the test from this PR may not detect it?), but for some reason, it couldn't be exported to Promethesus endpoint.
> > @douglas-reid
> 
> Without more information, I'm not sure how to respond. You definitely can't have a label name be the empty string. That shouldn't be possible in the other abstraction either.

Sorry if I wasn't clear. None of the tag key or values were empty.

douglas-reid:
> Sorry if I wasn't clear. None of the tag key or values were empty.

Do you have some example that you can share?  Can you construct a test case using this library that can be shared?

pitlv2109:
> > Sorry if I wasn't clear. None of the tag key or values were empty.
> 
> Do you have some example that you can share? Can you construct a test case using this library that can be shared?

https://gist.github.com/pitlv2109/d29b1b9dd819eb8b8fb814053b709521 

Run it, type something and check the endpoint `/metrics` on port 8888. You will see both key and values are there, but it'll show `Failed to export to Prometheus: "" is not a valid label name for metric "ocmetricstutorial_outgoing_latency"`. Thanks!

douglas-reid:
> https://gist.github.com/pitlv2109/d29b1b9dd819eb8b8fb814053b709521

Thanks for sending this!  Fix is available in https://github.com/istio/pkg/pull/32

pitlv2109:
Thanks for the quick fix :D

PR #34, opened 2019-07-16
Add a dockerfile image linter

fpesce:
This is a "companion PR" of istio/istio#15484
and istio/common-files#13


ozevren:
The file is not found. Do you need to update the common files?

geeknoid:
Note that this repo doesn't have any docker files and is unlikely to ever get any docker files, since the repo is intended to hold leaf utility packages. So this linter check won't ever be firing. It doesn't hurt to run it if you'd like to though...

PR #42, opened 2019-07-25
Retire now superfluous pkg/annotations.

geeknoid:


nmittler:
@geeknoid thanks for doing this! I got distracted :/

PR #46, opened 2019-07-31
Switch this repo to BUILD_WITH_CONTAINER=1

geeknoid:


Makefile
```diff
@@ -1,19 +1,19 @@
+img := gcr.io/istio-testing/build-tools:2019-07-31
```
sdake: woudl be nice to break out the HUB seperately as a followup to permit third parties to build their own ci in air-gapped mode.

du.txt
```diff
@@ -0,0 +1,14008 @@
+4	/boot
```
sdake: is this file intentional?

Makefile
```diff
@@ -1,19 +1,19 @@
...
+pwd := $(shell pwd)
+uid := $(shell id -u)
+run = docker run -t -i --sig-proxy=true --rm -v /etc/passwd:/etc/passwd:ro -v $(pwd):/work -w /work $(img)
+make = $(run) make -f Makefile.container.mk
+
 gen:
-	@go generate ./...
+	@$(make) gen
```
howardjohn: What do you think of making a wildcard rule the takes anything and runs it in the docker one? This may not be needed here but for Istio we have 100s of rules

geeknoid: Yeah, it was on my todo list to figure out how to do that. Make can do everything, so long as you know the secret language...

howardjohn:
I think it should be possible, I'll try it out tomorrow

Makefile.container.mk
```diff
@@ -0,0 +1,16 @@
+gen:
+	@go generate ./...
+
+lint:
+	@scripts/check_license.sh
+	@golangci-lint run -j 8 -v ./...
+
+fmt:
+	@goimports -w -local "istio.io" $(find . -type f -name '*.go' ! -name '*.gen.go' ! -name '*.pb.go' )
```
howardjohn: I think this should be `$(shell find ....)` not just find

Makefile
howardjohn:
```
%:
	$(make) $@
```

geeknoid: Excellent, thanks.

howardjohn: if you throw enough symbols at make something will work eventually :slightly_smiling_face:

geeknoid:
Note that the presubmit jobs are failing because the job definitions will need to be changed to run with the right container and trigger Makefile.container.mk instead of the normal Makefile.

Makefile
```diff
@@ -1,19 +1,12 @@
...
-
-test_with_coverage:
-	@go test -race -coverprofile=coverage.txt -covermode=atomic ./...
-	@curl -s https://codecov.io/bash | bash -s -- -c -F aFlag -f coverage.txt
+img := gcr.io/istio-testing/build-tools:2019-08-02
+pwd := $(shell pwd)
+uid := $(shell id -u)
+gocache := $(shell go env GOCACHE)
+gomod := $(GOPATH)/pkg/mod
+run = docker run -t -i --sig-proxy=true -u $(uid) --rm -v /etc/passwd:/etc/passwd:ro -v $(gocache):/gocache -v $(gomod):/go/pkg/mod -v $(pwd):/work -w /work $(img)
```
sdake: for this command - you might instead use ```run = docker run -t --sig-proxy=true -u $(uid) --rm \
+       -v /etc/passwd:/etc/passwd:ro \
+       --mount type=bind,source="$(gocache)",destination="/gocache",bind-propagation=rshared \
+       --mount type=bind,source="$(gomod)",destination="/go/pkg/mod",bind-propagation=rshared \
+       --mount type=bind,source="$(pwd)",destination="/work",bind-propagation=rshared \
+       -w /work $(img)
```

`-v` will create directories as root if they are not present.  This presents a problem in that the build will not proceed on a clean system.  `--mount` does not create directories and allows finer grained control over the mount operation.  It may be necessary to add some `mkdir -p` operations in here as well.

**Comment by geeknoid** [comment]
With the current container image, this mostly works. Two problems:

- When starting to Go compiler, I get:

   warning: GOPATH set to GOROOT (/go) has no effect

- When running "make lint", most of the linting works but then at the end there is:

INFO [lintersdb] Optimized sublinters [staticcheck gosimple unused stylecheck] into metalinter megacheck 
INFO [loader] Go packages loading at mode load deps types and syntax took 13.1927533s 
ERRO Running error: context loading failed: failed to load program with go/packages: go [list -e -json -compiled -test=true -export=false -deps=true -find=false -- ./...]: exit status 2: warning: GOPATH set to GOROOT (/go) has no effect
# runtime/cgo
exec: "gcc": executable file not found in $PATH
 
I think we need to include the gcc bits in the container for cgo's sake

**Comment by sdake** [comment]
ya the gcc toolchain isn't present - nor is it in the path if it were.

I'll take a look.  Groan - more crap in the container :)

**Comment by sdake** [comment]
might add a few `mkdir -p` operatons to the build:

```
sudo rm -rf /home/sdake/go/pkg/mod
sdake@beast-01:~/go/src/istio.io/pkg$ sudo rm -rf ~/.cache
sdake@beast-01:~/go/src/istio.io/pkg$ make build
docker: Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /home/sdake/.cache/go-build.
See 'docker run --help'.
Makefile:19: recipe for target 'build' failed
make: *** [build] Error 125
sdake@beast-01:~/go/src/istio.io/pkg$ mkdir /home/dake/.cache/go-build
mkdir: cannot create directory ‘/home/dake/.cache/go-build’: No such file or directory
sdake@beast-01:~/go/src/istio.io/pkg$ mkdir -p /home/sdake/.cache/go-build
sdake@beast-01:~/go/src/istio.io/pkg$ make build
docker: Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /home/sdake/go/pkg/mod.
See 'docker run --help'.
Makefile:19: recipe for target 'build' failed
make: *** [build] Error 125
sdake@beast-01:~/go/src/istio.io/pkg$ mkdir /home/sadke/go/pkg/mod
mkdir: cannot create directory ‘/home/sadke/go/pkg/mod’: No such file or directory
sdake@beast-01:~/go/src/istio.io/pkg$ mkdir -p /home/sdake/go/pkg/mod
```

File: Makefile [hunk 6]
```diff
@@ -1,19 +1,34 @@
...
+run = docker run -t --sig-proxy=true -u $(uid) --rm \
+       -v /etc/passwd:/etc/passwd:ro \
+       --mount type=bind,source="$(gocache)",destination="/gocache" \
+       --mount type=bind,source="$(gomod)",destination="/go/pkg/mod" \
+       --mount type=bind,source="$(pwd)",destination="/work" \
+       -w /work $(img)
 
-fmt:
-	@scripts/run_gofmt.sh
+ifdef CI
```
sdake: this part feels a little unnatural - detecting if running in ci...

Makefile
```diff
@@ -1,19 +1,34 @@
...
+endif
 
-test_with_coverage:
-	@go test -race -coverprofile=coverage.txt -covermode=atomic ./...
-	@curl -s https://codecov.io/bash | bash -s -- -c -F aFlag -f coverage.txt
+make = $(run) make -f Makefile.core.mk
 
+# default target
+build:
+	@$(make) build
```
sdake: this wont work if run = "".

PR #48, opened 2019-08-01
Add Viperize for brevity

therealmitchconnors:
Original functionality in this module had to be called three times (including a separate viper call) per command in order to provide consistent functionality.  This function need only be called once per root command, reducing the amount of code that has to be splattered across the codebase by about 9x.  Once Istio has been updated to use only viperize, we might consider making the other functions private.

### DIFF
--- a/viperconfig/viperconfig.go
+++ b/viperconfig/viperconfig.go
@@ -71,10 +71,7 @@ func subCommandPreRun(cmd *cobra.Command, viper *viper.Viper) {
 		if original != nil {
 			original(cmd, args)
 		}
-		flags := cmd.Flags()
-		pflags := cmd.PersistentFlags()
-		fmt.Println(pflags)
-		viper.BindPFlags(flags)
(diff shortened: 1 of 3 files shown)


viperconfig/viperconfig.go
```diff
@@ -56,3 +56,34 @@ func ProcessViperConfig(cmd *cobra.Command, viper *viper.Viper) {
...
+func subCommandPreRun(cmd *cobra.Command, viper *viper.Viper) {
+	original := cmd.PreRun
+	cmd.PreRun = func(cmd *cobra.Command, args []string) {
+		if original != nil {
+			original(cmd, args)
+		}
+		flags := cmd.Flags()
+		pflags := cmd.PersistentFlags()
+		fmt.Println(pflags)
+		viper.BindPFlags(flags)
```
golangcibot: Error return value of `viper.BindPFlags` is not checked (from `errcheck`)

PR #60, opened 2019-08-29
Remove Host, User, and Hub from version string

howardjohn:
These fields are not useful, and I don't think are appropriate to
include in our builds.

Host and User are either not useful or not "professional" - the normal
build will be "root" and some random SHA, but in a security fix it may
show up as howardjohn (this has happened in official releases).

Dockerhub doesn't really make any sense. When I run `make istioctl` it
doesn't have anything to do with docker. Even for pilot, which will
eventually be put into a dockerfile, it doesn't make sense. When the
binary is built we should only include metadata about that binary --
where it will eventually be packaged is just a guess that is wrong in
many cases.

Kubernetes was used as inspiration -- they do not have these fields.

For https://github.com/istio/istio/issues/16673

### DIFF
--- a/version/cobra_test.go
+++ b/version/cobra_test.go
@@ -107,15 +107,15 @@ func TestOpts(t *testing.T) {
 }
 
 var meshInfoSingleVersion = MeshInfo{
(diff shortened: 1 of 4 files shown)


howardjohn:
@geeknoid good point, ill send an update

PR #62, opened 2019-09-03
Add a build target and update the test target.

geeknoid:
Make this repo consistent with the others.


geeknoid:
Depends on https://github.com/istio/test-infra/pull/1754

howardjohn:
doesn't look like the codecov upload works, fails with  400?

PR #57, opened 2019-08-21
Fix data race in filewatcher.

ozevren:
Fixes https://github.com/istio/istio/issues/15987

PR #63, opened 2019-09-04
Reimplement FileWatcher without races.

geeknoid:
This reimplements the file watcher functionality to try and systematically eliminate the inherent race conditions in the previous design. There are some minor semantic changes:

- The FileWatcher.Add call now panics when called multiple times for the same path.

- The FileWatcher.Remove call now panics when called for an unknown path.

- It's now possible to watch a file that doesn't exist. Events will be generated when the file is created/deleted over time.

Code coverage was boosted to 98.3%.

This PR incorporates some of the changes from https://github.com/istio/pkg/pull/57
Fixes https://github.com/istio/istio/issues/15987

### DIFF
--- a/filewatcher/filewatcher.go
+++ b/filewatcher/filewatcher.go
@@ -1,4 +1,4 @@
-// Copyright 2018 Istio Authors
+// Copyright Istio Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


filewatcher/filewatcher.go
```diff
@@ -15,270 +15,180 @@
...
 
-	delete(worker.watchedFiles, cleanedPath)
-	close(tracker.events)
-	close(tracker.errors)
-	if len(worker.watchedFiles) == 0 {
-		// Remove the watch if all of its paths have been removed.
-		delete(w.workers, parentPath)
-		return worker.watcher.Close()
+	ws.count--
+	if ws.count == 0 {
```
ozevren: If you embed count operations into removePath/addPath (and perhaps make removePath return bool), then you can avoid changing a member field on worker, and avoid workerState struct.

PR #83, opened 2019-10-23
[release-1.4] add DockerInfo with hub and tag variables

carolynhu:


carolynhu:
@howardjohn can't agree more!

PR #97, opened 2020-01-05
Add probe cobra command

nak3:
mixer, galley and security(istio_ca) have `probe` sub-command and
their codes are duplicated.

This patch adds probe cobra command to be commonly used.

Part of https://github.com/istio/istio/issues/19922

### DIFF
--- a/probe/cobra.go
+++ b/probe/cobra.go
@@ -1,0 +1,35 @@
+package probe
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+func CobraCommand() *cobra.Command {
+	var (
+		probeOptions Options
+	)
+
+	prb := &cobra.Command{
+		Use:   "probe",
+		Short: "Check the liveness or readiness of a locally-running server",
+		Args:  cobra.ExactArgs(0),
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			if !probeOptions.IsValid() {
+				return fmt.Errorf("some options are not valid")
+			}
+			if err := NewFileClient(&probeOptions).GetStatus(); err != nil {
+				return fmt.Errorf("fail on inspecting path %s: %v", probeOptions.Path, err)
+			}
+			fmt.Fprintf(cmd.OutOrStdout(), "OK")
+			return nil
+		},
+	}
+	prb.PersistentFlags().StringVar(&probeOptions.Path, "probe-path", "",
+		"Path of the file for checking the availability.")
+	prb.PersistentFlags().DurationVar(&probeOptions.UpdateInterval, "interval", 0,
+		"Duration used for checking the target file's last modified time.")
+
+	return prb
+}
--- a/probe/cobra.go
+++ b/probe/cobra.go
@@ -1,3 +1,17 @@
+// Copyright 2020 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package probe
 
 import (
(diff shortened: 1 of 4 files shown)


probe/cobra.go
```diff
@@ -0,0 +1,35 @@
...
+		probeOptions Options
+	)
+
+	prb := &cobra.Command{
+		Use:   "probe",
+		Short: "Check the liveness or readiness of a locally-running server",
+		Args:  cobra.ExactArgs(0),
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			if !probeOptions.IsValid() {
+				return fmt.Errorf("some options are not valid")
```
esnible: This error message is poor.  Can you change IsValid() to return an `error` or add `func (o *Options) Validate() error` so that a real message can be returned?

probe/cobra.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+func CobraCommand() *cobra.Command {
+	var (
+		probeOptions Options
+	)
+
+	prb := &cobra.Command{
+		Use:   "probe",
+		Short: "Check the liveness or readiness of a locally-running server",
```
esnible: Can you add an `Example:` section?  Some of the options, like `probe-path`, are not clear to me.

probe/cobra.go
```diff
@@ -0,0 +1,35 @@
...
+				return fmt.Errorf("some options are not valid")
+			}
+			if err := NewFileClient(&probeOptions).GetStatus(); err != nil {
+				return fmt.Errorf("fail on inspecting path %s: %v", probeOptions.Path, err)
+			}
+			fmt.Fprintf(cmd.OutOrStdout(), "OK")
+			return nil
+		},
+	}
+	prb.PersistentFlags().StringVar(&probeOptions.Path, "probe-path", "",
```
esnible: I don't know what this means.  Can you rewrite the message so someone unfamiliar will be able to understand?  I realize you are just moving this code but perhaps now is the time to improve the message.

probe/cobra.go
```diff
@@ -0,0 +1,35 @@
...
+		"Path of the file for checking the availability.")
+	prb.PersistentFlags().DurationVar(&probeOptions.UpdateInterval, "interval", 0,
+		"Duration used for checking the target file's last modified time.")
```
esnible: Can you rewrite this or ask for someone from @istio/wg-docs-maintainers to take a look.  I don't understand what this means.  Will `mixer probe` fail if the file is older than the interval?

probe/cobra.go
```diff
@@ -0,0 +1,35 @@
+package probe
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+func CobraCommand() *cobra.Command {
```
esnible: I don't know if we `lint` istio/pkg but can you put a comment in front of every exported function so that we can generate docs?

nak3: Done.

probe/cobra.go
nak3: Sure, added `Validate()`. Done.

probe/cobra.go
nak3: Done.

probe/cobra.go
nak3: Done.

probe/cobra.go
nak3: Yes, you are correct. `probe` checks the last modified time and it fails if it is older than the interval in [GetStatus()](https://github.com/istio/pkg/blob/master/probe/client.go#L40). It may be needless to say, but server side like `mixer server` updates the file's timestamp.

IMO, it would be better to add more detail in the help description so I added it https://github.com/istio/pkg/pull/97/files#diff-d50c947d5e646a9af72cb90ca19920ccR12-R13 . If I should put more detail in the option, please let me know (I am pinging docs team in docs channel in the Slack to take a look as well.)

nak3:
Thank you @esnible.

The help option after the updates prints it like this:

```
$  ./linux_amd64/galley probe -h
Check the liveness or readiness of a locally-running server

Check if the target file is older than the interval. The locally-running server updates the last modifiled time
(you need to enable it on server side), so probe command compares it with current time.

Usage:
  galley probe [flags]

Examples:

  # Check if the target file '/health' is older than interval of 4 seconds
  probe --probe-path=/health --interval=4s

Flags:
  -h, --help                help for probe
      --interval duration   Duration used for checking the target file's last modified time.
      --probe-path string   Path of the file for checking the last modified time.

Global Flags:
  ... snip ...
```

probe/cobra.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+var (
+	probeLongDesc = `Check the liveness or readiness of a locally-running server
+
+Check if the target file is older than the interval. The locally-running server updates the last modifiled time
```
ericvn: c/modifiled/modified

nak3: Done.

esnible:
@ericvn I like what I am seeing.  My review isn't enough; you will need to re-review.

PR #112, opened 2020-02-24
add log_utc_time_zone flag for utc time zone

xc8801:
Issue Desc:
1.  time.Now() // local time zone
2. t.UTC() // set utc time zone
3. log output utc time zone and can't use local time zone.

fix:
1. add log_utc_time_zone flag.
2. --log_utc_time_zone=true // default &&  set UTC time zone.
3. --log_utc_time_zone=false // set location time zone.

### DIFF
--- a/log/config.go
+++ b/log/config.go
@@ -167,7 +167,6 @@ func prepZap(options *Options) (zapcore.Core, zapcore.Core, zapcore.WriteSyncer,
 }
 
 func formatDate(t time.Time, enc zapcore.PrimitiveArrayEncoder) {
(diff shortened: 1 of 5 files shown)


xc8801:
@howardjohn 
We are in Asia, The problem we are facing now is that the time when the log system appears between the UTC time zone and the local time zone is different , so our local time zone is strongly need to be supported . May be add one flag, to avoid break the current default rule of UTC and control UTC enable status.

howardjohn:
Envoy does not support this, and we need our logging to be consistent with Envoy

PR #120, opened 2020-03-07
Fix istioctl controlz dashboard UI shows error message in  subpage

AsCat:
Some debug info was removed in the dashboard UI template, which causes error when user using "istioctl dashboard controlz" command.

The related commit is [here](https://github.com/istio/pkg/commit/8f434ef6237f2b722d38f5d04f3c9bca5602f9fe#diff-22d6d38cd847736f6c99f3133279067c)

Related isuue i posted is [here](https://github.com/istio/istio/issues/21959)

### DIFF
--- a/ctrlz/topics/assets/assets.gen.go
+++ b/ctrlz/topics/assets/assets.gen.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"time"
 )
+
 type asset struct {
 	bytes []byte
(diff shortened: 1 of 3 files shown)


esnible:
@AsCat This PR was approved, but it isn't building.  Do you think merging with master would help?

AsCat:
@esnible I fixed the build error, and now it's merged, thanks for your remind.

PR #142, opened 2020-04-20
Add hook to record method in monitoring pkg

bianpengyuan:
Add per measure record hook, so that vendor logic could be executed when record is called.

### DIFF
--- a/monitoring/monitoring.go
+++ b/monitoring/monitoring.go
@@ -17,6 +17,7 @@ package monitoring
 import (
 	"context"
 	"fmt"
+	"sync"
 
 	"go.opencensus.io/stats"
 	"go.opencensus.io/stats/view"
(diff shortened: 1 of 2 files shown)


bianpengyuan:
@douglas-reid Do you mean adding an example like https://github.com/istio/pkg/blob/949b673891de95fdd7941f541019f6a9335547da/monitoring/example_sum_test.go? Use case being vendor could transform OSS measure into desired measure in custom build by registering hook logic.

douglas-reid:
@bianpengyuan is this a workaround for providing a custom view ?

bianpengyuan:
I think we will need custom measure as well. This hook is to enable that.

douglas-reid:
For custom measures, can't we use normal mechanisms?

bianpengyuan:
That means we will need to add a redundant line of code to increment the vendor measure wherever the OSS measure gets incremented, which is going to spread over the code base. The vendor patch will consistently be rebased on top of OSS history, so it will conflict a lot if we do that. While with this hook, we could modularize vendor measure increment logic, which to me is cleaner and easier to maintain. Anyway, this is just one way that I can think of, and I don't feel strong about it as well since it introduces some runtime cost to transform the measurement. 

If there is a better way to achieve "increment measurement a when increment measurement b" while avoiding widespread code change that would be great.

PR #192, opened 2020-07-19
Error dictionary and structured logging

ostromart:
Implementation of https://docs.google.com/document/d/1vdYswLQuYnrLA2fDjk6OoZx2flBABa18UjCGTn8gsg8.

### DIFF
--- a/errdict/errdict.go
+++ b/errdict/errdict.go
@@ -2,11 +2,17 @@ package errdict
 
 import "istio.io/pkg/monitoring"
 
+// IstioErrorStruct represents structured error information, for optional use in scope.X or log.X calls.
+// See https://docs.google.com/document/d/1vdYswLQuYnrLA2fDjk6OoZx2flBABa18UjCGTn8gsg8/ for additional information.
 type IstioErrorStruct struct {
-	MoreInfo     string
-	Impact       string
-	Action       string
-	LikelyCauses string
+	// MoreInfo is additional information about the error e.g. a link to context describing the context for the error.
+	MoreInfo string
+	// Impact is the likely impact of the error on system function e.g. "Proxies are unable to communicate with Istiod."
+	Impact string
+	// Action is the next step the user should take e.g. "Open an issue or bug report."
+	Action string
+	// LikelyCause is the likely cause for the error e.g. "Software bug."
+	LikelyCause string
 
 	// Metrics is a list of associated metrics that will be passed to a handler from the call site.
 	Metrics []monitoring.Metric
--- a/errdict/errdict.go
+++ b/errdict/errdict.go
@@ -1,0 +1,13 @@
+package errdict
+
+import "istio.io/pkg/monitoring"
+
+type IstioErrorStruct struct {
+	MoreInfo     string
+	Impact       string
+	Action       string
+	LikelyCauses string
+
+	// Metrics is a list of associated metrics that will be passed to a handler from the call site.
+	Metrics []monitoring.Metric
+}
--- a/go.mod
+++ b/go.mod
@@ -25,6 +25,7 @@ require (
 	github.com/onsi/gomega v1.5.0
 	github.com/pkg/errors v0.8.1 // indirect
 	github.com/prometheus/client_golang v0.9.4
+	github.com/prometheus/common v0.4.1
 	github.com/prometheus/procfs v0.0.8 // indirect
 	github.com/prometheus/prom2json v1.1.0
 	github.com/spaolacci/murmur3 v0.0.0-20180118202830-f09979ecbc72
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,6 @@ require (
 	github.com/onsi/gomega v1.5.0
 	github.com/pkg/errors v0.8.1 // indirect
 	github.com/prometheus/client_golang v0.9.4
-	github.com/prometheus/common v0.4.1
 	github.com/prometheus/procfs v0.0.8 // indirect
 	github.com/prometheus/prom2json v1.1.0
 	github.com/spaolacci/murmur3 v0.0.0-20180118202830-f09979ecbc72
--- a/log/scope.go
+++ b/log/scope.go
@@ -28,12 +28,11 @@ import (
 // Scope constrains logging control to a named scope level. It gives users a fine grained control over output severity
 // threshold and stack traces.
 //
-// Scope supports structured logging using WithLabels and WithoutLabels:
+// Scope supports structured logging using WithLabels:
 //
 //   s := RegisterScope("MyScope", "Description", 0)
 //   s = s.WithLabels("foo", "bar", "baz", 123, "qux", 0.123)
 //   s.Info("Hello")                      // <time>   info   MyScope   Hello  foo=bar baz=123 qux=0.123
-//   s.WithoutLabels("qux").Info("Hello") // <time>   info   MyScope   Hello  foo=bar baz=123
 //
 // The output format can be globally configured to be JSON instead, using Options in this package.
 //  e.g. <time>   info   MyScope   { "message":"Hello","foo":"bar","baz":123 }
@@ -348,8 +347,8 @@ func (s *Scope) GetLogCallers() bool {
 	return s.logCallers.Load().(bool)
 }
 
-// Copy makes a copy of s and returns a pointer to it.
-func (s *Scope) Copy() *Scope {
+// copy makes a copy of s and returns a pointer to it.
+func (s *Scope) copy() *Scope {
 	out := *s
 	out.labels = copyStringInterfaceMap(s.labels)
 	return &out
@@ -359,7 +358,7 @@ func (s *Scope) Copy() *Scope {
 // It returns a copy of s, with the labels added.
 // e.g. newScope := oldScope.WithLabels("foo", "bar", "baz", 123, "qux", 0.123)
 func (s *Scope) WithLabels(kvlist ...interface{}) *Scope {
-	out := s.Copy()
+	out := s.copy()
 	if len(kvlist)%2 != 0 {
 		out.labels["WithLabels error"] = fmt.Sprintf("even number of parameters required, got %d", len(kvlist))
 		return out
@@ -375,31 +374,6 @@ func (s *Scope) WithLabels(kvlist ...interface{}) *Scope {
 		out.labels[key] = kvlist[i+1]
 		out.labelKeys = append(out.labelKeys, key)
 	}
-	return out
-}
-
-// WithoutLabels makes a copy of s, clears labels in s with the given keys and returns the copy.
-// Not-existent keys are ignored.
-// e.g. myScope = myScope.WithLabels("foo", "bar", "baz", 123, "qux", 0.123)
-//      myScope = myScope.WithoutLabels("baz", "qux")
-// Result: myScope retains only the "foo" label.
-func (s *Scope) WithoutLabels(keys ...string) *Scope {
-	out := s.Copy()
-	for _, key := range keys {
-		delete(out.labels, key)
-		out.labelKeys = removeKey(out.labelKeys, key)
-	}
-	return out
-}
-
-// WithoutAnyLabels clears all labels from a copy of s and returns it.
-// e.g. myScope = myScope.WithLabels("foo", "bar", "baz", 123, "qux", 0.123)
-//      myScope = myScope.WithoutAnyLabels()
-// Result: myScope has no labels.
-func (s *Scope) WithoutAnyLabels() *Scope {
-	out := s.Copy()
-	out.labels = make(map[string]interface{})
-	out.labelKeys = nil
 	return out
 }
 
@@ -436,14 +410,4 @@ func copyStringInterfaceMap(m map[string]interface{}) map[string]interface{} {
 		out[k] = v
 	}
 	return out
-}
-
-func removeKey(s []string, key string) []string {
-	for i, k := range s {
-		if k == key {
-			s = append(s[:i], s[i+1:]...)
-			return s
-		}
-	}
-	return s
 }
--- a/log/scope.go
+++ b/log/scope.go
@@ -25,8 +25,27 @@ import (
 	"istio.io/pkg/structured"
 )
 
-// Scope let's you log data for an area of code, enabling the user full control over
-// the level of logging output produced.
+// Scope constrains logging control to a named scope level. It gives users a fine grained control over output severity
+// threshold and stack traces.
+//
+// Scope supports structured logging using WithLabels and WithoutLabels:
+//
+//   s := RegisterScope("MyScope", "Description", 0)
+//   s = s.WithLabels("foo", "bar", "baz", 123, "qux", 0.123)
+//   s.Info("Hello")                      // <time>   info   MyScope   Hello  foo=bar baz=123 qux=0.123
+//   s.WithoutLabels("qux").Info("Hello") // <time>   info   MyScope   Hello  foo=bar baz=123
+//
+// The output format can be globally configured to be JSON instead, using Options in this package.
+//  e.g. <time>   info   MyScope   { "message":"Hello","foo":"bar","baz":123 }
+//
+// Scope also supports an error dictionary. The caller can pass a *structured.Error object as the first parameter
+// to any of the output functions (Fatal*, Error* etc.) and this will append the fields in the object to the output:
+//
+//   e := &structured.Error{MoreInfo:"See the documentation in istio.io/helpful_link"}
+//   s.WithLabels("foo", "bar").Error(e, "Hello")
+//     <time>   info   MyScope   Hello  moreInfo=See the documentation in istio.io/helpful_link foo=bar
+//
+// See structured.Error for additional guidance on defining errors in a dictionary.
 type Scope struct {
 	// immutable, set at creation
 	name        string
@@ -60,7 +79,7 @@ type scopeHandlerCallbackFunc func(
 	fields []zapcore.Field)
 
 // registerDefaultHandler registers a scope handler that is called by default from all scopes. It is appended to the
-// current list of scope handlers.
+// current list of default scope handlers.
 func registerDefaultHandler(callback scopeHandlerCallbackFunc) {
 	lock.Lock()
 	defer lock.Unlock()
@@ -336,7 +355,9 @@ func (s *Scope) Copy() *Scope {
 	return &out
 }
 
-// WithLabels adds a key-value pairs to the labels in s.
+// WithLabels adds a key-value pairs to the labels in s. The key must be a string, while the value may be any type.
+// It returns a copy of s, with the labels added.
+// e.g. newScope := oldScope.WithLabels("foo", "bar", "baz", 123, "qux", 0.123)
 func (s *Scope) WithLabels(kvlist ...interface{}) *Scope {
 	out := s.Copy()
 	if len(kvlist)%2 != 0 {
@@ -359,6 +380,9 @@ func (s *Scope) WithLabels(kvlist ...interface{}) *Scope {
 
 // WithoutLabels makes a copy of s, clears labels in s with the given keys and returns the copy.
 // Not-existent keys are ignored.
+// e.g. myScope = myScope.WithLabels("foo", "bar", "baz", 123, "qux", 0.123)
+//      myScope = myScope.WithoutLabels("baz", "qux")
+// Result: myScope retains only the "foo" label.
 func (s *Scope) WithoutLabels(keys ...string) *Scope {
 	out := s.Copy()
 	for _, key := range keys {
@@ -369,6 +393,9 @@ func (s *Scope) WithoutLabels(keys ...string) *Scope {
 }
 
 // WithoutAnyLabels clears all labels from a copy of s and returns it.
+// e.g. myScope = myScope.WithLabels("foo", "bar", "baz", 123, "qux", 0.123)
+//      myScope = myScope.WithoutAnyLabels()
+// Result: myScope has no labels.
 func (s *Scope) WithoutAnyLabels() *Scope {
 	out := s.Copy()
 	out.labels = make(map[string]interface{})
--- a/log/scope.go
+++ b/log/scope.go
@@ -16,14 +16,13 @@ package log
 
 import (
 	"fmt"
-	"runtime"
 	"strings"
 	"sync"
 	"sync/atomic"
-	"time"
 
-	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
+
+	"istio.io/pkg/errdict"
 )
 
 // Scope let's you log data for an area of code, enabling the user full control over
@@ -39,10 +38,38 @@ type Scope struct {
 	outputLevel     atomic.Value
 	stackTraceLevel atomic.Value
 	logCallers      atomic.Value
+
+	// context data - key slice to preserve ordering
+	contextKey []string
+	context    map[string]interface{}
+	// structured data - key slice to preserve ordering
+	structuredKey []string
+	structured    map[string]interface{}
+	// scope specific handlers
+	handlers []ScopeHandlerCallbackFunc
+	mu       sync.RWMutex
 }
 
 var scopes = make(map[string]*Scope)
 var lock = sync.RWMutex{}
+
+type ScopeHandlerCallbackFunc func(
+	level Level,
+	scope *Scope,
+	ie *errdict.IstioErrorStruct,
+	msg string,
+	fields []zapcore.Field)
+
+var (
+	defaultHandlers []ScopeHandlerCallbackFunc
+	mu              sync.RWMutex
+)
+
+func RegisterDefaultHandler(callback ScopeHandlerCallbackFunc) {
+	mu.Lock()
+	defer mu.Unlock()
+	defaultHandlers = append(defaultHandlers, callback)
+}
 
 // RegisterScope registers a new logging scope. If the same name is used multiple times
 // for a single process, the same Scope struct is returned.
@@ -74,6 +101,9 @@ func RegisterScope(name string, description string, callerSkip int) *Scope {
 		scopes[name] = s
 	}
 
+	s.context = make(map[string]interface{})
+	s.structured = make(map[string]interface{})
+
 	return s
 }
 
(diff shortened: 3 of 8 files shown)


log/scope_test.go
```diff
@@ -218,6 +220,130 @@ func TestBasicScopes(t *testing.T) {
...
+		s.Debuga("Hello")
+		s2.Debuga("Hello")
+
+		_ = Sync()
+	})
+	if err != nil {
+		t.Errorf("Got error '%v', expected success", err)
+	}
+
+	mustRegexMatchString(t, lines[0], `Hello {"foo":"bar","baz":123}`)
```
howardjohn, line 309: Are there logging ingestion tools that read logs that are partially plaintext and partially json?

log/zapcore_handler.go
```diff
@@ -0,0 +1,129 @@
...
+func ZapLogHandlerCallbackFunc(
+	level Level,
+	scope *Scope,
+	ie *errdict.IstioErrorStruct,
+	msg string,
+	fields []zapcore.Field) {
+	var ss []string
+	if ie != nil {
+		ss = append(ss, fmt.Sprintf(`"moreInfo":"%s"`, ie.MoreInfo))
+		ss = append(ss, fmt.Sprintf(`"impact":"%s"`, ie.Impact))
```
howardjohn, line 48: hand crafting json with arbitrary inputs means the output may easily become invalid json

log/zapcore_handler.go
```diff
@@ -0,0 +1,129 @@
+package log
+
+import (
+	"fmt"
+	"runtime"
+	"strings"
+	"time"
+
+	"github.com/prometheus/common/log"
```
howardjohn, line 9: Where does this come from?

log/zapcore_handler.go
```diff
@@ -0,0 +1,129 @@
...
+
+func toZapSlice(index int, fields ...interface{}) []zapcore.Field {
+	var zfs []zapcore.Field
+	if len(fields) <= index {
+		return nil
+	}
+	for _, zfi := range fields {
+		zf, ok := zfi.(zapcore.Field)
+		if !ok {
+			log.Errorf("bad interface type: expect zapcore.Field, got %T for fields %v", zf, fields)
```
howardjohn, line 79: This should definitely not be here (it is the Prometheus log library)

log/scope_test.go
```diff
@@ -218,6 +220,130 @@ func TestBasicScopes(t *testing.T) {
...
+		funcs.Store(funcs.Load().(patchTable))
+		s.WithLabel("foo", "bar").Debuga(ie, "Hello")
+
+		_ = Sync()
+	})
+	if err != nil {
+		t.Errorf("Got error '%v', expected success", err)
+	}
+
+	mustRegexMatchString(t, lines[0], `{"moreInfo":"MoreInfo","impact":"Impact","action":"Action","likelyCauses":"LikelyCause","foo":"bar","message":"Hello"}`)
```
howardjohn, line 335: Why is message last? It seems the most important

log/scope_test.go
```diff
@@ -218,6 +220,130 @@ func TestBasicScopes(t *testing.T) {
 	}
 }
 
+func TestScopeAddLabel(t *testing.T) {
+	const name = "TestScope"
```
howardjohn, line 224: I am very confused how I am supposed to use the library now. Seems like we have 3 distinct ways to add metadata (scope, label, context) with no clear distinction as to when I should use one or the other

log/scope_test.go
```diff
@@ -218,6 +220,130 @@ func TestBasicScopes(t *testing.T) {
...
+		_ = Sync()
+	})
+
+	if err != nil {
+		t.Errorf("Got error '%v', expected success", err)
+	}
+
+	expect := `Hello {"foo":"bar","baz":1,"qux":0.123}`
+	mustRegexMatchString(t, lines[0], expect)
+	mustRegexMatchString(t, lines[1], expect)
```
howardjohn, line 242: why is there 2 lines?
Why is it a regex match instead of exact match - from the timestamp?

log/scope.go
```diff
@@ -315,3 +343,149 @@ func (s *Scope) SetLogCallers(logCallers bool) {
...
+	ief, ok := fields[0].([]interface{})
+	if !ok {
+		return nil, 0
+	}
+	ie, ok := ief[0].(*errdict.IstioErrorStruct)
+	if !ok {
+		return nil, 0
+	}
+	// Skip IstioErrorStruct, pass remaining fields on as before.
+	return ie, 1
```
douglas-reid, line 472: clarifying: to use the structured error, it *must* be the first argument in the `Debuga` (or similar) call?  Is the intent to make the API the same while adding the functionality, then?

Is adding a bunch of `DebugStructured()`, etc.,  that enforce the ordering directly problematic because of all the levels ?

log/scope_test.go
douglas-reid, line 335: i'm a little nervous that the output changes format based on the inclusion of the `ie` in the call to `Debuga()`. That seems like it might surprise.

Somewhat related, I don't know what the right thing here is, but I worry that we'll have alternating log lines like:

```text
Hello {"foo": "bar", "baz": "goo"}
{"moreInfo"... "message":"Hello"}
Goodbye {"foo"...}
{"moreInfo"..."message":"Goodbye"}
```

This seems potentially like a mess to parse. Am I reading this incorrectly?

errdict/errdict.go
```diff
@@ -0,0 +1,19 @@
+package errdict
+
+import "istio.io/pkg/monitoring"
+
+// IstioErrorStruct represents structured error information, for optional use in scope.X or log.X calls.
+// See https://docs.google.com/document/d/1vdYswLQuYnrLA2fDjk6OoZx2flBABa18UjCGTn8gsg8/ for additional information.
+type IstioErrorStruct struct {
```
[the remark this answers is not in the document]
ostromart: All of those are better than IstioErrorStruct - I vote for structured.Error. btw. the reason for the Struct at the end was that I was planning to wrap this in an interface, but then I decided the extra complexity is not worth it.

errdict/errdict.go
```diff
@@ -0,0 +1,19 @@
...
+	// MoreInfo is additional information about the error e.g. a link to context describing the context for the error.
+	MoreInfo string
```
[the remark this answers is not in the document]
ostromart: It's in the original message text itself i.e. what we would print today. We clearly need more docs and examples here but I wanted to get basic consensus before generating a ton of docs for this.

errdict/errdict.go
```diff
@@ -0,0 +1,19 @@
...
+	// Impact is the likely impact of the error on system function e.g. "Proxies are unable to communicate with Istiod."
+	Impact string
+	// Action is the next step the user should take e.g. "Open an issue or bug report."
+	Action string
+	// LikelyCause is the likely cause for the error e.g. "Software bug."
+	LikelyCause string
```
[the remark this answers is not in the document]
ostromart: Not yet, but the idea would be to have some consts defined in the dictionary package. I also considered having this as a slice but it gets too unwieldly.

errdict/errdict.go
```diff
@@ -0,0 +1,19 @@
...
+
+	// Metrics is a list of associated metrics that will be passed to a handler from the call site.
+	Metrics []monitoring.Metric
```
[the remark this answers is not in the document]
ostromart: There's an example/description in the doc and I will copy it here if that looks good?

log/scope.go
```diff
@@ -315,3 +343,149 @@ func (s *Scope) SetLogCallers(logCallers bool) {
 func (s *Scope) GetLogCallers() bool {
 	return s.logCallers.Load().(bool)
 }
+
+// RegisterHandler registers a handler for s that is called for all Fatal*, Error*, Warn*, Info* and Debug* function calls.
+func (s *Scope) RegisterHandler(callback ScopeHandlerCallbackFunc) {
```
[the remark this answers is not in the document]
ostromart: It's meant to be used to pass the same structured info (labels, context) primarily also to a metrics handler in the future. There could also be a couple of different flavor of text loggers (e.g. performance vs. nice output).
 I think there's very low likelihood of anyone except us adding any handlers but we could make this private interface for now.

log/scope.go
```diff
@@ -315,3 +343,149 @@ func (s *Scope) SetLogCallers(logCallers bool) {
...
+func (s *Scope) AddContext(key string, value interface{}) *Scope {
+	mu.Lock()
+	defer mu.Unlock()
+	s.context[key] = value
+	s.contextKey = append(s.contextKey, key)
+	return s
+}
+
+// WithContext returns a copy of s with the given key-value appended to the context.
+func (s *Scope) WithContext(key string, value interface{}) *Scope {
```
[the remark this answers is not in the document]
ostromart: See the unit tests. Add* means it persists for future scope calls, With* is for this call alone.

log/scope.go
```diff
@@ -315,3 +343,149 @@ func (s *Scope) SetLogCallers(logCallers bool) {
...
+func (s *Scope) ClearAllContext() *Scope {
+	mu.Lock()
+	defer mu.Unlock()
+	s.context = make(map[string]interface{})
+	s.contextKey = nil
+	return s
+}
+
+// AddLabel adds a key-value pair to the labels in s.
+func (s *Scope) AddLabel(key string, value interface{}) *Scope {
```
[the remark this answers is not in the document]
ostromart: context just adds some OT compliant formatting and it's meant for things like transaction ID. not sure that we need both though, they are very similar in output - thoughts @douglas-reid ?

log/scope.go
ostromart: Right, exactly. An alternative would be to name the calls something different (InfoStructf, DebugStructf...). 
I'm ok with either approach. The second approach would be a brand new thing rather than trying for backward compatibility which could actually be a good thing.

log/scope_test.go
ostromart: See comment about label/context - we can remove one of these if it's going to be too confusing when to use one or the other. 
I think the difference between scope and label/context should be clear though - labels/context are bound to a scope. I agree more examples/docs are needed.

log/scope_test.go
```diff
@@ -218,6 +220,130 @@ func TestBasicScopes(t *testing.T) {
...
+func TestScopeAddLabel(t *testing.T) {
+	const name = "TestScope"
+	const desc = "Desc"
+	s := RegisterScope(name, desc, 0)
+	s.SetOutputLevel(DebugLevel)
+
+	lines, err := captureStdout(func() {
+		funcs.Store(funcs.Load().(patchTable))
+		s.AddLabel("foo", "bar").AddLabel("baz", 123).AddLabel("qux", 0.123).Debuga("Hello")
+		s.Debuga("Hello")
```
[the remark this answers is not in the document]
ostromart: WithLabel would not apply to either call, only if we had s.WithLabel().Info(). 
We need something that controls "for this debug statement only" vs. "add to the scope for every subsequent debug statement" - any suggestions how that should look? I don't think assignment can be used to trigger afaik in Go.

log/scope_test.go
ostromart: Yes, but it's up to us to decide exactly how the log outputs. It could be all JSON, text + JSON or all text. That can easily be made more configurable.

log/scope_test.go
ostromart: Good point, how about:
```
Hello {"foo": "bar", "baz": "goo"}
Hello { "message":"Hello","moreInfo"...}
```

log/zapcore_handler.go
```diff
@@ -0,0 +1,129 @@
...
+// support for errdict package and structured logging.
+func ZapLogHandlerCallbackFunc(
+	level Level,
+	scope *Scope,
+	ie *errdict.IstioErrorStruct,
+	msg string,
+	fields []zapcore.Field) {
+	var ss []string
+	if ie != nil {
+		ss = append(ss, fmt.Sprintf(`"moreInfo":"%s"`, ie.MoreInfo))
```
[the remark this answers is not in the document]
ostromart: this will be configurable. the plan was to add that in a follow-on PR but i will include it in here.

log/zapcore_handler.go
ostromart: i think we can do better here to avoid that. i'd rather not include a json library to format the json though. 
how about, number types have no quotes, everything else has escaped quotes? then objects will not cause invalid json, only messy json :-)

log/zapcore_handler.go
```diff
@@ -0,0 +1,129 @@
...
+		ss = append(ss, fmt.Sprintf(`"impact":"%s"`, ie.Impact))
+		ss = append(ss, fmt.Sprintf(`"action":"%s"`, ie.Action))
+		ss = append(ss, fmt.Sprintf(`"likelyCauses":"%s"`, ie.LikelyCause))
+	}
+	if len(scope.structuredKey) > 0 {
+		scope.mu.RLock()
```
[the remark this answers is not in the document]
ostromart: ack, per offline discussion this was to avoid copies but if copies are ok we can get rid of this. i will update with the locks removed and a few more copies.

log/zapcore_handler.go
```diff
@@ -0,0 +1,129 @@
...
+				ss = append(ss, fmt.Sprintf(`"%s":"%v"`, k, v))
+			} else {
+				ss = append(ss, fmt.Sprintf(`"%s":%v`, k, v))
+			}
+		}
+		scope.mu.RUnlock()
+	}
+	if len(ss) > 0 {
+		ss = append(ss, fmt.Sprintf(`"message":"%s"`, msg))
+		msg = "{" + strings.Join(ss, ",") + "}"
```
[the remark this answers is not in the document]
ostromart: i think it does warrant some explanation at the file comment level. zap structured logging is not much more than what is here and it seems pointless to conform this to zap structures just to leverage that tiny bit of code. the structured logging is actually not more performant - the lean and mean log implementation is in the "emit" call which we still retain.

log/zapcore_handler.go
```diff
@@ -0,0 +1,129 @@
...
+		if !ok {
+			log.Errorf("bad interface type: expect zapcore.Field, got %T for fields %v", zf, fields)
+			continue
+		}
+		zfs = append(zfs, zf)
+	}
+	return zfs
+}
+
+const callerSkipOffset = 4
```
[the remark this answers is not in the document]
ostromart: there wasn't in the original code i got it from but i will add one.

log/zapcore_handler.go
```diff
@@ -0,0 +1,129 @@
...
+	pt := funcs.Load().(patchTable)
+	if pt.write != nil {
+		if err := pt.write(e, fields); err != nil {
+			_, _ = fmt.Fprintf(pt.errorSink, "%v log write error: %v\n", time.Now(), err)
+			_ = pt.errorSink.Sync()
+		}
+	}
+}
+
+// backwards compatibility. TODO(mostrowski): remove this
```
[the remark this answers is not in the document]
ostromart: it's to preserve the semantics in the default.go file. i'll get rid of it shortly.

log/scope_test.go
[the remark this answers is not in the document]
howardjohn: I would suggest always copying. If you want to persist the changes you do `scope = scope.WithLabels()`. This seems more explicit/idiomatic, and then we only need a single function?

log/zapcore_handler.go
ostromart: from goland being helpful :-)

log/scope_test.go
douglas-reid: i'm fairly torn on this. what does zap default to?  does the console encoder thingy use half text, half JSON?  if so, then we should probably match that, and then only switch to full JSON if that's requested in the config (which I believe is your proposal).

As an aside, I almost feel like cli-style (`Hello foo=bar baz=goo`) or full JSON (`{"message":"Hello", "foo":"bar", ...}`) are better than mixing JSON and free text, but that seems like an issue for zap, not Istio.

log/scope.go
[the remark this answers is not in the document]
ostromart: removed this in favor of copying.

log/scope.go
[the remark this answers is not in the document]
ostromart: removed the scope version. let's stick with the global handlers to limit the scope for bad things to happen.

log/zapcore_handler.go
ostromart: removed.

log/scope_test.go
ostromart: removed context for now. we can discuss what to do with it in a chat with doug.

log/zapcore_handler.go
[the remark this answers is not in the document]
ostromart: i took john's suggestion to just use zap. since zap is all about performance i trust all their type reflection is blazing fast :-)

log/scope_test.go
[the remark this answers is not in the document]
ostromart: ok, agreed. i was worried about having too many copies but this is way simpler and in reality i think the copies are not a problem.

ostromart:
@howardjohn @douglas-reid ptal a look again, i've incorporated most of your suggestions:
 - removed context (for now, we can put it back)
 - all scope functions return a copy so i could remove the locking and the semantics are simpler now
 - refactored the structured output to use zap fields (and hence their JSON encoder)
 - renamed errdict.IstioErrorStruct to structured.Error
 - updated the field ordering to put the message first
 - message is always output outside the JSON block for consistency. 
 - message is the first item in the JSON block
if you guys are ok with the results i'll start work on documentation.

go.mod
```diff
@@ -25,6 +25,7 @@ require (
 	github.com/onsi/gomega v1.5.0
 	github.com/pkg/errors v0.8.1 // indirect
 	github.com/prometheus/client_golang v0.9.4
+	github.com/prometheus/common v0.4.1
```
howardjohn, line 28: this is suspicious - hopefully the test fails

log/scope.go
```diff
@@ -39,10 +38,37 @@ type Scope struct {
 	outputLevel     atomic.Value
 	stackTraceLevel atomic.Value
 	logCallers      atomic.Value
+
+	// context data - key slice to preserve ordering
+	contextKey []string
+	context    map[string]interface{}
```
howardjohn, line 44: I thought we got rid of context?

log/scope.go
```diff
@@ -39,10 +38,37 @@ type Scope struct {
...
-var lock = sync.RWMutex{}
+var (
+	scopes          = make(map[string]*Scope)
+	defaultHandlers []ScopeHandlerCallbackFunc
+	lock            sync.RWMutex
+)
+
+// ScopeHandlerCallbackFunc is a callback type for the handler called from Fatal*, Error*, Warn*, Info* and Debug*
+// function calls.
+type ScopeHandlerCallbackFunc func(
```
howardjohn, line 58: nit: private type?

log/scope.go
```diff
@@ -39,10 +38,37 @@ type Scope struct {
...
+	level Level,
+	scope *Scope,
+	ie *structured.Error,
+	msg string,
+	fields []zapcore.Field)
+
+// RegisterDefaultHandler registers a scope handler that is called by default from all scopes. It is appended to the
+// current list of scope handlers.
+func RegisterDefaultHandler(callback ScopeHandlerCallbackFunc) {
```
howardjohn, line 67: nit: private function?

howardjohn: I care a decent amount about making everything in log that can be private, private. To avoid confusion

log/scope.go
```diff
@@ -130,27 +165,33 @@ func (s *Scope) FatalEnabled() bool {
...
 	if s.GetOutputLevel() >= ErrorLevel {
-		s.emit(zapcore.ErrorLevel, s.GetStackTraceLevel() >= ErrorLevel, fmt.Sprint(args...), nil)
+		ie, firstIdx := getErrorStruct(args)
+		s.callHandlers(ErrorLevel, s, ie, fmt.Sprint(args[firstIdx:]...), nil)
 	}
 }
 
 // Errorf uses fmt.Sprintf to construct and log a message at error level.
-func (s *Scope) Errorf(template string, args ...interface{}) {
+// func (s *Scope) Errorf(template string, args ...interface{}) {
```
howardjohn, line 186: nit: stray comments

log/scope.go
```diff
@@ -130,27 +165,33 @@ func (s *Scope) FatalEnabled() bool {
...
+func (s *Scope) Errorf(args ...interface{}) {
```
howardjohn, line 187: This *might* be a backwards incompatible API change. Or maybe not. But why do we need it? Is it valid to pass `Errorf(<not string>, something)`? What happens in this case?

log/scope.go
```diff
@@ -315,3 +338,96 @@ func (s *Scope) SetLogCallers(logCallers bool) {
...
+	msg string,
+	fields []zapcore.Field) {
+	for _, h := range defaultHandlers {
+		h(severity, scope, ie, msg, fields)
+	}
+}
+
+// getErrorStruct returns (*Error, 1) if it is the first argument in the list is an Error ptr,
+// or (nil,0) otherwise. The second return value is the offset to the first non-Error field.
+func getErrorStruct(fields ...interface{}) (*structured.Error, int) {
```
howardjohn, line 404: nit: why even return int here? seems like callers can just check if the response is == nil. Its internal though so NBD

log/scope_test.go
howardjohn: how come in the previous test its json with spaces and in this one there are no spaces?

log/scope_test.go
```diff
@@ -218,6 +220,89 @@ func TestBasicScopes(t *testing.T) {
...
+		funcs.Store(funcs.Load().(patchTable))
+		s.WithLabels("foo", "bar").Debuga(ie, "Hello")
+
+		_ = Sync()
+	})
+	if err != nil {
+		t.Errorf("Got error '%v', expected success", err)
+	}
+
+	mustRegexMatchString(t, lines[0], `Hello	{"message": "Hello", "moreInfo": "MoreInfo", "impact": "Impact", "action": "Action", "likelyCauses": "LikelyCause", "foo": "bar"}`)
```
howardjohn, line 294: Where does the big space between Hello and `{` come from? Or is a tab and it displays weird on the diff?

howardjohn, line 294: Do we need the message in two places? Seems a bit redundant. I guess for log processors they just look at the json part?

log/zapcore_handler.go
```diff
@@ -0,0 +1,123 @@
...
+// ZapLogHandlerCallbackFunc is the handler function that emulates the previous Istio logging output and adds
+// support for errdict package and labels logging.
+func ZapLogHandlerCallbackFunc(
+	level Level,
+	scope *Scope,
+	ie *structured.Error,
+	msg string,
+	fields []zapcore.Field) {
+	if ie != nil {
+		fields = append(fields, zap.String("message", msg))
```
howardjohn, line 44: We currently have a `--log_as_json` option. What happens when this is enabled and we have structured logging? hopefully not double nested json?

howardjohn: As a further point, does it make sense to output json only if that flag is enabled, and if its disabled something like `Hello moreInfo=some more info impact=some impact action=some action likelyCauses=some cause`? Similar to https://github.com/sirupsen/logrus format I guess

log/zapcore_handler.go
```diff
@@ -0,0 +1,123 @@
...
+		fields = append(fields, zap.String("moreInfo", ie.MoreInfo))
+		fields = append(fields, zap.String("impact", ie.Impact))
```
howardjohn, line 46: should we append only if they are not empty?

log/scope.go
```diff
@@ -315,3 +338,96 @@ func (s *Scope) SetLogCallers(logCallers bool) {
...
+	out.context = copyStringInterfaceMap(s.context)
+	out.labels = copyStringInterfaceMap(s.labels)
+	return &out
+}
+
+// WithLabels adds a key-value pairs to the labels in s.
+func (s *Scope) WithLabels(kvlist ...interface{}) *Scope {
+	out := s.Copy()
+	if len(kvlist)%2 != 0 {
+		out.labels["WithLabels error"] = fmt.Sprintf("even number of parameters required, got %d", len(kvlist))
```
howardjohn, line 354: Just curious what `zap` does here? I can't think of a better way to do things, just curious if someone from zap has

go.mod
ostromart: done.

log/scope.go
ostromart: forgot to remove this bit, done.

log/scope.go
ostromart: good idea, done.

log/scope.go
ostromart: done.

log/scope.go
ostromart: done.

log/scope.go
ostromart: we need it to retain the familiar interface to scope.Debug* et al. while allowing structured.Error to be optionally passed. the alternative (per thread with doug) is to call the functions taking structured.Error something else. but what? i feel it could get messy to have a proliferation of these functions with slightly different names.

log/scope.go
ostromart: just to avoid checking every time at the caller site. makes the code more compact for the 50 wrapper functions.

log/scope_test.go
ostromart: these are both due to switching to the zap json encoder. it's their fault now!

ostromart: per doug's point, we need to be consistent. but we can debate exactly what format we want to emit and it's easy to change it. if we vote for either cli or json and not a mix i'm ok with that.

log/zapcore_handler.go
ostromart: yes, will make this change in a bit.

log/zapcore_handler.go
ostromart: done.

log/scope.go
ostromart: they use fprintf :-D

log/zapcore_handler.go
ostromart: done.

log/scope_test.go
ostromart: fyi. zap does a mix of tabbed and json (for the fields) but i agree with doug that this is inconsistent.

structured/structured.go
```diff
@@ -0,0 +1,35 @@
...
+	MoreInfo string
+	// Impact is the likely impact of the error on system function e.g. "Proxies are unable to communicate with Istiod."
+	Impact string
+	// Action is the next step the user should take e.g. "Open an issue or bug report."
+	Action string
+	// LikelyCause is the likely cause for the error e.g. "Software bug."
+	LikelyCause string
+
+	// Metrics is a list of associated metrics that will be passed to a handler from the call site.
+	Metrics []monitoring.Metric
```
douglas-reid, line 34: since the `Metrics` part doesn't appear in any tests, and isn't used yet, should we remove it from the initial PR in the goal of starting simple, or do we feel like it will be used in short order?

ostromart: unfortunately the deadline is today so there wouldn't be time to do it separately. yes, i will put up some samples pretty quickly afterwards (cleaning logs and adding metrics shouldn't be subject to code freeze).

log/scope.go
```diff
@@ -100,27 +144,30 @@ func Scopes() map[string]*Scope {
...
+func (s *Scope) Fatal(fields ...interface{}) {
 	if s.GetOutputLevel() >= FatalLevel {
-		s.emit(zapcore.FatalLevel, s.GetStackTraceLevel() >= FatalLevel, msg, fields)
+		ie, firstIdx := getErrorStruct(fields)
+		s.callHandlers(FatalLevel, s, ie, fields[0].(string), toZapSlice(firstIdx+1, fields))
 	}
 }
 
 // Fatala uses fmt.Sprint to construct and log a message at fatal level.
 func (s *Scope) Fatala(args ...interface{}) {
```
howardjohn, line 155: I am not sure how this differs from `Fatal` in terms of output. I think with your interface changes it may actually be identical? Can we verify and if so as a followup remove the `*a` ones?

log/scope.go
```diff
@@ -315,3 +347,103 @@ func (s *Scope) SetLogCallers(logCallers bool) {
 func (s *Scope) GetLogCallers() bool {
 	return s.logCallers.Load().(bool)
 }
+
+// Copy makes a copy of s and returns a pointer to it.
+func (s *Scope) Copy() *Scope {
```
howardjohn, line 352: Is there a use case for this or should it be private?

log/scope.go
```diff
@@ -315,3 +347,103 @@ func (s *Scope) SetLogCallers(logCallers bool) {
...
+	}
+	return out
+}
+
+// WithoutLabels makes a copy of s, clears labels in s with the given keys and returns the copy.
+// Not-existent keys are ignored.
+// e.g. myScope = myScope.WithLabels("foo", "bar", "baz", 123, "qux", 0.123)
+//      myScope = myScope.WithoutLabels("baz", "qux")
+// Result: myScope retains only the "foo" label.
+func (s *Scope) WithoutLabels(keys ...string) *Scope {
```
howardjohn, line 386: nit: do we need this? I would expect usage to be

```
s := some-scope
if ... {
  s = s.WithLabels(foo)
  s.Log(something with foo)
}
s.Log(something without foo)
```

howardjohn: Basically I am biased in favor of minimal API until proven a need otherwise. We can always add, hard to remove and use a big API though. But no strong opinion here

log/scope.go
howardjohn: ack, thanks

log/scope.go
ostromart: Fatala doesn't take structured parameters to pass to zap. If no one uses those then we can remove (actually I'd replace Fatal with the Fatala implementation and remove Fatala).

log/scope.go
ostromart: yes, thanks for pointing out, it should be now.

log/scope.go
ostromart: that's a good question. i guess i took it for granted that someone would want to do this but when pressed i'm having a hard time coming up with a convincing example. i will remove for now, we can indeed easily add it back.

PR #215, opened 2020-08-09
Add structured.Error return through error type handling

ostromart:
This allows called functions to return structured.Error in place of error, with correct rendering when scope.X is called with the returned error.


structured/structured.go
```diff
@@ -27,4 +29,28 @@ type Error struct {
...
+	// Err is the original error string.
+	Err error
+}
+
+// Error implements the error#Error interface.
+func (e *Error) Error() string {
+	if e == nil {
+		return ""
+	}
+	ret := fmt.Sprintf("\tmoreInfo=%s impact=%s action=%s likelyCause=%s",
```
howardjohn, line 41: when is this function actually used? it seems like it will only be used if some accidentally uses the logging library wrong? Is there another use? It seems weird we prepend tab as well

richardwxn: Is it possible that the original error string is nil, if so then makes sense

structured/structured.go
```diff
@@ -27,4 +29,28 @@ type Error struct {
...
+		e.MoreInfo, e.Impact, e.Action, e.LikelyCause)
```
richardwxn, line 42: Is it possible that the original error string is nil, if so then makes sense

howardjohn (dismissed):
The more I think about it the less I think this makes sense - but I don't have the context you have. Consider the case of something like

```go
err := someFuncReturningStructuredError()
if err != nil {
  log.Error(structured.Error{err, MoreInfo: foobar})
}
```

This will end up logging something like `moreInfo=foobar err= moreInfo=some more info`. And even worse, if we have json format, it will be a mix of json and non json. 

Also what happens if I log the error like `Errorf("%v", structured)` instead of `Error(structured)`? Now it *almost* looks right, but it doesn't call the hooks, record metrics, work with json mode, etc.

What is the motivation for this? I can see how it could make sense but I am not sure I understand how yet

structured/structured.go
ostromart: it would be used when someone does something like this:
```
return fmt.Errorf("some info: %v", structured.Error)
```
i.e. annotate an error as it's going up the stack. it's fairly common. in that case, at that point the structured type would unavoidably get rendered into a regular error. 
the reason for the tab is to maintain consistency with the regular formatting (without the flattening) which puts a tab between the message and the structured section.

structured/structured.go
ostromart: yes, it's possible if the user returns a structured.Error type without the Err field populated.

ostromart:
the use case is returning a structured.Error from functions that would normally return error type. it does depend on the dev using it correctly, otherwise the formatting could indeed be off. the way to use it is something like this:
```
iff err != nil {
  return structured.NewErr(myStructuredError, err)
}
```
as in the test case. it works with labels too. in that case everything is rendered correctly, even if this return is deep in the stack and the next function up the stack does something like this: 
```
return fmt.Errorf("annotation: %v", structured.Error)
```
which would flatten the error. the way it works is that the rendering is done through the Error() method at this point rather than in the handler.
an alternative way to handle this would be to unflatten the message string in the hander and reconstitute the rendered fields from the message string. it's certainly a viable approach but it does come with a.performance penalty. and there's no way that i can think of to prevent the structured type being flattened when it's passed up the stack with an annotation like above.

howardjohn:
> it does depend on the dev using it correctly

It doesn't depend on the dev using it correctly, it depends on leaky abstractions... if a function returns `errror`, I should be able to use it however I use any error. If I *must* use it a certain way, don't pretend its an `error`. Its not just the format, its also the hooks and metrics

ostromart:
i'm very much open to suggestions here - i don't immediately see a better way, but that doesn't mean there isn't one. 
most functions we have return error and i think it's infeasible to refactor all the code to use some internal error type.  there are lots of places where the source of the error is only known at the original returned error, not where scope.X is actually called. structured logging (labels) is currently a function of scope, not structured.Error so to support labels that would need to be added to structured.Error (i.e. func (e *Error) WithLabels() etc).
but fundamentally the question is whether to return through an error type. i really think the only option is yes. if so, we need some way to serialize/deserialize structured.Error into an error type.

howardjohn:
IMO we should only be using this stuff for logging. The entire API is designed around this, its fundamentally broken to use it in the way you are doing here. Only the caller has enough context to determine things like "likely causes" or give proper context for an error. In my opinion, it is the functions responsibility is to return an error like `failed to find virtual service`, not a massive struct about metrics, contacting support, calling `foobar` hook. If the caller wants to then log this with a bunch of info, they can break out the full struct. If they need to propagate more info up the stack, they can do so using standard error. only when we are logging do we add the full structure.

Put another way, `error` is for robots, `structured.Error` is for humans (logging).

ostromart:
> Only the caller has enough context to determine things like "likely causes" or give proper context for an error.

if that were really true i'd agree this is not necessary, but in almost all cases there's a huge fanout to functions that generate the error and all get logged through an entry in the top level function, which has absolutely no idea what the real problem could be. the called is typically just logging a message about a potentially huge variety of errors coming up the stack. and both logging + returning an error is a go anti-pattern.

howardjohn:
If we want to do that then we need to solve:
* incorrect formatting (json vs text)
* misleading/unused hooks/metrics
* Not logging the same ~message~ paragraph about contacting support 10 times when an error goes up 10 call stacks
* Not break things like `errors.IsNotFound` (for k8s client)

ostromart:
> incorrect formatting (json vs text)
agreed, this is just work, nothing fundamental preventing this.

> misleading/unused hooks/metrics
it would not be unused - i think the best way forward is to deserialize back to structured.Error before calling the hooks. then it would work exactly as before.

> Not logging the same message paragraph about contacting support 10 times when an error goes up 10 call stacks
not sure i follow this point. where are we logging anything ten times when it goes up the stack? i'm not aware of such cases but if we are doing that it's wrong and should be fixed.

> Not break things like errors.IsNotFound (for k8s client)
it will not break. the go error wrapping has its own scheme for (de)serializing. the metadata i assume is encoded in a sufficiently robust way that all sorts of random text in the message should not confuse it.

howardjohn:
> agreed, this is just work, nothing fundamental preventing this.

I don't actually see how its possible... if I run `fmt.Errorf("some context: %v", structredError)` we lose all contexts

> is to deserialize back to structured.Error 

How? all we have is a string

>  where are we logging anything ten times when it goes up the stack? 

Say I call `fmt.Errorf(structured.Error(fmt.Errorf("context: %v": someerror))`, where someerror is also a strutured error (which I don't know - its just an `error`). Now I will get two structured outputs. If both of these have the same boilerplate about contacting support, for example, we have tons of duplicates. This is one example - if we have a deep call tree we could presumably have the same "contact support" paragraph logged 10x. It will literally fill up the entire console

ostromart:
to answer both questions above, we'd (de)serialize it into the error string itself (just as grpc.Error does). the format would be human readable as a string for compatibility with error but woudl have some sentinel or format that allows us to easily deserialize. i'll take a look at the exact implementation in the grpc package but from memory it's just a special string format.
 the case of decorating a structured.Error with another structured.Error as it goes up the stack is a 1% scenario - the meat of the error is almost always generated at the original error site in Istio and as it goes up the stack it's going to be decorated with text  not another structured.Error. If anything, i think we should actually clean up cases where the decorations add no value and just do return err passthrough. However, a simple way to handle that would be just to discard the structured part and overwrite with the last structured.Error content (except the original error text that is, we'd follow the regular error style chaining for that) as we go up the stack. it could also be nested but i don't believe this situation is common enough to warrant implementing that.

ostromart:
@howardjohn I took another pass at making this work. The deserialization now looks for a regex in the serialized string if enclosing markers are found. I think it's sufficiently performant and robust, and still human readable if for some reason the string is not logged through a scope.

log/scope_test.go
```diff
@@ -283,13 +284,54 @@ func TestScopeErrorDictionary(t *testing.T) {
...
+
 		_ = Sync()
 	})
 	if err != nil {
 		t.Errorf("Got error '%v', expected success", err)
 	}
 
-	mustRegexMatchString(t, lines[0], `Hello	moreInfo=MoreInfo impact=Impact action=Action likelyCauses=LikelyCause foo=bar`)
+	mustRegexMatchString(t, lines[0], `Hello	moreInfo=MoreInfo impact=Impact action=Action likelyCause=LikelyCause 	foo=bar`)
+	mustRegexMatchString(t, lines[1], `Hello: 	moreInfo=MoreInfo impact=Impact action=Action likelyCause=LikelyCause err=err 	foo=bar`)
```
howardjohn, line 300: This is still destroying the intent of `"Hello: %s"` producing a standard go error message of `context: some more context: some more context`. Especially in json output, we end up with 
```json
{
  "level": "debug",
  "time": "2020-09-17T18:46:18.149655Z",
  "scope": "TestScope",
  "msg": "",
  "message": "Hello: ",
  "moreInfo": "MoreInfo",
  "impact": "Impact",
  "action": "Action",
  "likelyCause": "LikelyCause",
  "err": "err",
  "foo": "bar"
}
```

which doesn't seem to make much sense to me

howardjohn (commented):
I think I just disagree this is a good idea. I don't have any suggestions to make it better, because I don't think we should do it at all. I appreciate the problem you are trying to solve, but I have not seen a similar pattern in any other projects, and I think it will lead to a lot of confusion and issues. See https://github.com/istio/pkg/pull/215#pullrequestreview-463925878 and my other comments for rationale - some have been addressed but many still remain, especially the ones around the metrics, etc. I think we are mixing logging and errors in dangerous and unexpected ways.

I will remove my hold, so if others feel this is a good idea we can go ahead with it.

log/scope_test.go
ostromart: What do you expect to see with JSON instead of the above?

howardjohn: That is my main point - I don't think there *is* an output that makes sense, because I think the idea of mixing logging and `error` is not the right approach

ostromart:
When you say you haven't seen a similar pattern, how is https://godoc.org/google.golang.org/grpc/status different from this?

howardjohn:
grpc status .Error() function is returning a simple string `return fmt.Sprintf("rpc error: code = %s desc = %s", codes.Code(e.e.GetCode()), e.e.GetMessage())` that we don't try to do some parsing hacks inside of our logging library, and it doesn't try to add metrics, callbacks, etc into logging.

grpc.status is just a normal error as far as I can tell, nothing weird there. If I do `log.Errorf("foo: %v", structuredError)` and I don't get a message of the form `foo: ....` then we have violated expectations

log/scope_test.go
ostromart: I'm confused... the CLI output above makes sense but there's no reasonable JSON representation of the same?
If moreInfo and the other fields were just labels and not coming from structured.Error, would the output be ok to you then?

ostromart:
the parts which are the same:
- grpc and this both implement Error interface method so you can return additional info through error. there's literally no difference in the implementation here (both return a string with the extra fields rendered).
- both also have FromError which extracts the fields. grpc does it through a typecast rather than string, which means that once you return it through fmt.Errorf("annotation: %v", err), FromError no longer works.

i haven't seen any example where the logger would inherently do something unexpected: log.Errorf("foo: %v", structuredError) would output foo: &{...} just as it would with any struct ptr. i agree that annotations should stack as we go up the stack: "annotation2: annotation1: original error". can you give an example of where this wouldn't happen with this logger? that's how this is supposed to work anyway but i may be missing something.
the reason i am pushing for it is that most (80%?) structured errors come from somewhere down the stack and will get flattened along the way. the place that does the logging in most cases has no context at all, it just usually does log.Errorf("Logging some error: %s", err) where err comes from somewhere deep in the stack with all the context. so realistically there's few places you can insert a structured.Error at this point.
the alternative is to log at the site, which i would argue is even more messy (you'll definitely get duplicate messages then) and a go antipattern.
i have a proposal which may do some way to addressing your concerns and simplifying things. i want to do this anyway but haven't had time yet. since we're essentially returning the equivalent of a code, we can make Error() just output a code with a string so the log messages are "annotation2: annotation1: error message (see istio.io/errors/123456)".  the urls would be generated from code. 
i still think we should unflatten the string in that case though to pass a structured.Error to the handler. i think the typecast method is too unreliable. or not unflatten and give up on the idea of having general handlers that consume structured.Error.

ostromart:
This has gotten bogged down so trying a hopefully less controversial approach in https://github.com/istio/pkg/pull/236.

PR #219, opened 2020-08-27
Refactor log functions to use Scope logging methods

tbarrella:
Enables passing an error struct as part of [logging improvements](https://docs.google.com/document/d/1vdYswLQuYnrLA2fDjk6OoZx2flBABa18UjCGTn8gsg8/edit)

### DIFF
--- a/log/default_test.go
+++ b/log/default_test.go
@@ -18,6 +18,8 @@ import (
 	"regexp"
 	"strconv"
 	"testing"
+
+	"istio.io/pkg/structured"
 )
 
 func testOptions() *Options {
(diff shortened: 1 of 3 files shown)


nmittler:
I'm actually a little surprised this wasn't what we were already doing :)

Assuming the actual log methods are checking for the log level, then this should be exactly the same. Otherwise, there may be an impact on performance.

log/default_test.go
```diff
@@ -215,6 +217,32 @@ func TestDefault(t *testing.T) {
...
+
+		Infoa(ie, "Hello")
+		_ = Sync()
+	})
+
+	if err != nil {
+		t.Errorf("Got error '%v', expected success", err)
+	}
+
+	expected := "Hello	moreInfo=MoreInfo impact=Impact action=Action likelyCauses=LikelyCause"
```
ostromart, line 240: would be good to add WithLabels to this test.

PR #231, opened 2020-09-18
Pipe klog/v2 output through standard logging

howardjohn:
Fixes https://github.com/istio/istio/issues/26953

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/fsnotify/fsnotify v1.4.7
 	github.com/ghodss/yaml v1.0.0
+	github.com/go-logr/logr v0.2.0
 	github.com/gogo/protobuf v1.2.2-0.20190730201129-28a6bbf47e48 // indirect
 	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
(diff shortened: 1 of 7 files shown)


ostromart:
why do we want to support another logging api rather than consolidating to scope?

howardjohn:
@ostromart this is not about supporting another logging API for consumers of pkg/log. The stuff is all private. The key code is `klog.SetLogger(newLogrAdapter(KlogScope))`. Without this, kubernetes logging goes through its own logging system which makes istioctl print out *tons* of logs and not respect our logging options or formats. This makes it so we pipe the output into zap.

Previously, with klog v1 we had a fork in istio/klog. The proper way of doing this, per the klog docs, is to use SetLogger. SetLogger takes a logr.Logger interface, which is a standard interface for logging (without an implementation). There is a zap logr implementation but it also doesn't respect "Scope", since we invented our own thing, so I had to make our own.

ostromart:
ok, could you add a few more notes about why this is needed and how it's meant to be used? without that it's very hard to know what this is for and why the logging threshold is 3.

howardjohn:
@ostromart added docs about what I commented

ostromart (approved):

ericvn (approved):

PR #234, opened 2020-09-21
Fix zapcore field handling

ostromart:


log/scope_test.go
```diff
@@ -54,8 +55,8 @@ func TestBasicScopes(t *testing.T) {
 		},
 
 		{
-			f:   func() { s.Info("Hello") },
-			pat: timePattern + "\tinfo\ttestScope\tHello",
+			f:   func() { s.Info("Hello", zap.String("key", "value"), zap.Int("intkey", 123)) },
```
howardjohn, line 58: what about `Info("foo", "bar", "baz")`?

howardjohn: Seems like this should either work or the type should be changed on Info to only take ZapField?

ostromart: added, thanks

howardjohn: so its an error? then why don't we have the old `Info(msg string, fields ...zapcore.Field)` as the interface?

howardjohn: Well I guess the first maybe is not `msg string` if we allow the structured.Error - but seems the rest doesn't need to allow anything?

ostromart: i made it error because someone there are some use cases now of zap fields. but i can refactor those to the new structured api and make info() and others behave like fmt.sprint and forget about zap field support. 
regarding the second point, the first field is optional so the start index of the zap fields is not fixed.

howardjohn: My question is more about why does the interface allow passing anything but then error. Can we not have type safety here? It seems like we used to have it without issue.

ostromart: we cannot statically type it for the reason above i.e. the first field is optional.

howardjohn: So the current interface is `Info(fields ...interface{})`

Valid options are:
* Info(str)
* Info(str, zap.Field, zap.Field, ...)
* Info(structured.Error, zp.Field, zap.Field, ..)

Is that accurate?

if so, can't the interface be:

`Info(msgOrError interface{}, fields ...zapcore.Field)`

ostromart: the other valid option is Info(structured.Error, str, zp.Field, zap.Field, ..)

ostromart: how about the option of removing the zapcore fields altogether? there are not many uses and the structured logging in scope is what we want people to use.

howardjohn:
I see.. can we at least document the valid inputs then if we cannot put it
in the type?

log/scope_test.go
howardjohn: 100% fine with that

ostromart:
latest commit deprecates zap.Field support. if this looks ok @howardjohn please lgtm and i will refactor istio/istio code before this merges to remove use of zap.Field there.

howardjohn (approved):

howardjohn:
Looks good!

ostromart:
Needs https://github.com/istio/istio/pull/27477 to merge.

PR #236, opened 2020-09-22
Add support for wrapping structured.Error with %w

ostromart:
This PR implements %w wrapping for cases where structured.Error needs to be returned as an error from a function. 
See https://blog.golang.org/go1.13-errors for background on %w. 
This redoes the work in https://github.com/istio/pkg/pull/215 with some important differences:
- no special parsing required to extract fields for handlers
- caller is responsible for explicitly extracting structured.Error before calling scope.

### DIFF
--- a/structured/structured.go
+++ b/structured/structured.go
@@ -14,6 +14,10 @@
 
 package structured
 
+import (
+	"fmt"
+)
+
 // Error represents structured error information, for optional use in scope.X or log.X calls.
 // It is not the same thing as structured logging. The "structured" here means adding a structure to user facing
 // messages.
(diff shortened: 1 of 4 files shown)


structured/structured.go
```diff
@@ -27,4 +31,26 @@ type Error struct {
...
+// NewErr creates a new copy of an Error with the content of serr and err and returns a ptr to it.
+func NewErr(serr *Error, err error) *Error {
+	// Make a copy so that dictionary entry is not modified.
+	ne := *serr
+	ne.Err = err
+	return &ne
+}
+
+// Unwrap implements error unwrapping for %w verb.
+func (e *Error) Unwrap() error { return e }
```
howardjohn, line 56: Shouldn't this return `e.Err`? that is my understanding from https://blog.golang.org/go1.13-errors "The Unwrap method"

ostromart: i don't think so - in our use case we want to recover the original structured.Error rather than some partial representation of it. if you look at the implementation of errors.As(), it does a type match for the passed in receiver using Unwrap to get any errors embedded in the chain. since we are calling As() with a structured.Error received, the Unwrap method must return the same type.

howardjohn: I am not sure I understand. If we are doing this we don't need wrap is my understanding from "Examining errors with Is and As" example 2 - we can just do type casting. the point of unwrap is to expose the wrapped error.

Note - I barely new about %w until today, so not experienced here

ostromart: i'm no expert either, i also found out about it fairly recently so this is hardly authoritative, just how i understand it.
we still need %w for returns with annotations. if you just typecast in the caller frame, if anyone did return fmt.Errorf("foo: %v", err) in the stack, the typecast wouldn't work anymore. basically %w allows you to annotate by stashing the type in a private field when %w is used. the deficiency with this (as some blogs have noted) is that you _must_ use %w - if you accidentally use %v, the type is not recoverable. i hope that linters will pick this up...
btw. this is the advantage of encoding in the string - it will work even with %v - but i agree there's an ickyness to doing that. as long as linters pick this up, i think %w is the way to go.

howardjohn (approved):

PR #266, opened 2020-11-04
WIP: Add feature status to environment vars

therealmitchconnors:
This adds the concept of feature support for our various environment variables, and sets all variables to unknown for now.  Users can call SetVarStatus to update the support level of a given env var.

### DIFF
--- a/env/var.go
+++ b/env/var.go
@@ -17,6 +17,7 @@
 package env
 
 import (
+	"fmt"
 	"os"
 	"sort"
 	"strconv"
@@ -42,6 +43,27 @@ const (
 	DURATION
 )
 
+type FeatureStatus byte
+
+const (
+	Alpha FeatureStatus = iota
+	Beta
+	Stable
+	Unknown
+)
+
+func (s FeatureStatus) String() string {
+	switch s {
+	case Alpha:
+		return "Alpha"
+	case Beta:
+		return "Beta"
+	case Stable:
+		return "Stable"
+	}
+	return "Unknown"
+}
+
 // Var describes a single environment variable
 type Var struct {
 	// The name of the environment variable.
@@ -61,6 +83,9 @@ type Var struct {
 
 	// The type of the variable's value
 	Type VarType
+
+	// The support level of the feature controlled by this env var
+	FeatureStatus FeatureStatus
 }
 
 // StringVar represents a single string environment variable.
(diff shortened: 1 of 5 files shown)


env/var.go
```diff
@@ -42,6 +43,27 @@ const (
...
+)
+
+func (s FeatureStatus) String() string {
+	switch s {
+	case Alpha:
+		return "Alpha"
+	case Beta:
+		return "Beta"
+	case Stable:
+		return "Stable"
```
howardjohn, line 62: What does it mean for an env var to be stable? Don't we currently document that all env vars are alpha?

Put another way, why would we promote an env var to stable vs moving it to an API/mesh config?

therealmitchconnors: I am all for the idea of migrating all stable environment variables to an API, but that does not represent the current state of the project.  For instance, I believe we set the PILOT_TRACE_SAMPLING variable on every istio deployment, and it should probably be considered stable.

howardjohn: PILOT_TRACE_SAMPLING will be deprecated in 1.9 by the tracing api. We shouldn't mark something "stable" just because its been around for a while, we should mark it stable if we know its the right long term API and its properly tested

therealmitchconnors: The overall idea here is to give users a way to visualize the maturity of the features they are leveraging.  We've already got that for Annotations with Nate's PR, and this would provide visibility into env vars, which could be combined in a single istioctl command to show users what features they are using, and what support they can expect.  Given the prominence of many env vars, I doubt we would feed comfortable telling our users that all of them have only alpha level support.

env/var.go
```diff
@@ -42,6 +43,27 @@ const (
 	DURATION
 )
 
+type FeatureStatus byte
+
+const (
+	Alpha FeatureStatus = iota
```
louiscryan, line 49: In a world where we canonicalize these Unknown's should become forbidden. Can we add a 'Dev' status

env/var.go
louiscryan: Can we track the same state we're now tracking for annotation to cover the transition intent

deprecated, deprecatedBy

etc.

louiscryan:
What format does the ctrlz page take? Hopefully the same one we will use for the checked-in canonicalization.

therealmitchconnors:
> What format does the ctrlz page take? Hopefully the same one we will use for the checked-in canonicalization.

You can see the format here: https://drive.google.com/file/d/1EU9f711kP4TR6ixttN7GFohVcw48AGwR/view?usp=sharing

env/var.go
therealmitchconnors: Platform env vars are not registered with the env package, and the ctrlz page lists their default values as UNKNOWN.  I don't think we have any concept of 'Dev' status elsewhere in the project.  Would 'experimental' make more sense?

env/var.go
therealmitchconnors: Yes.  It might be helpful to track the removal of deprecated features too, so we can highlight that when checking upgrade readiness.

It's also useful to note that users may not be setting these directly, but via operator or helm values.

louiscryan:
In the UI it says the 'Set of environment variables defined for the process" which implies we're dumping everything. Does it make sense to be more selective? Options include

- Only show the environment variables that are 'read'
- Sort the display to show non-UNKNOWN variables first


WDYT?

therealmitchconnors:
The env vars are already sorted to show known env vars first.  I am happy to move the env vars which are not known to the ctrlz system (and therefore have no default) into a separate table if that helps.

env/var.go
linsun: Should we add experimental to here as well?

louiscryan:
What do we want the ctrlz page to show by default?

- All possible env vars
- All 'set' env vars
- All 'set' env vars with a non-default value

I suspect the latter would be the most useful. Variety of presentations
possible to achieve delineation (sorting, split tables etc.)

PR #278, opened 2020-11-18
Refactor Ledger for readability, allow erasing history

therealmitchconnors:
This refactor is primarily for readability, but also adds the ability to erase nodes that are no longer needed because newer versions have been globally distributed.  This refactor also results in a ~140% improvement on throughput:

```
BenchmarkScale
Before:
BenchmarkScale-12    	   17766	     68073 ns/op	   0.12 MB/s	   34039 B/op	     453 allocs/op
After:
BenchmarkScale-12    	   46752	     27500 ns/op	   0.29 MB/s	   14632 B/op	     237 allocs/op
```

This PR is WIP until it is documented, tested in Pilot, and I decide what to do with regard to history entries that are never sent to envoys.


ledger/ledger.go
```diff
@@ -40,73 +45,145 @@ type Ledger interface {
...
+	// this map will grow forever.  how do we limit it?
+	keyCache byteCache
 }
 
 // Make returns a Ledger which will retain previous nodes after they are deleted.
-func Make(retention time.Duration) Ledger {
-	return smtLedger{tree: newSMT(hasher, nil, retention)}
+// the retention parameter has been removed in favor of EraseRootHash, but is left
+// here for backwards compatibility
+func Make(_ time.Duration) Ledger {
```
howardjohn, line 68: So cache.NewTTL is not used now? I saw a few tests are still calling it, but in istio/istio we will not?

ledger/ledger.go
```diff
@@ -40,73 +45,145 @@ type Ledger interface {
...
+	return &smtLedger{
+		tree:    newSMT(hasher, nil),
```
howardjohn, line 70: nit: it seems newSMT makes `cache.NewTTL(forever, time.Second)`. This seems like a waste of a goroutine and it polls every 1s for something that will never happen (assuming `forever` doesn't occur :slightly_smiling_face: ). Its not made worse by this PR though as far as I can tell.

ledger/ledger.go
```diff
@@ -40,73 +45,145 @@ type Ledger interface {
...
+	err := s.tree.Erase(occurrences[0].Value.([]byte), adjacentRoots)
+	if err != nil {
+		return err
+	}
+	for _, o := range occurrences {
+		s.history.Remove(o)
+	}
+	s.history.lock.Lock()
+	defer s.history.lock.Unlock()
+	delete(s.history.index, rootHash)
```
howardjohn, line 96: nit: this seems like it should be a method on history rather than directly accessed

howardjohn:
I recommend using https://godoc.org/golang.org/x/perf/cmd/benchstat for comparing benchmarks, super useful

ledger/node.go
```diff
@@ -0,0 +1,307 @@
...
+		n.val[hashLength] = 0
+		n.page.nodes[leftIndex(n.index)] = nil
+		n.page.nodes[rightIndex(n.index)] = nil
+	}
+}
+
+func (n *node) calculateHash(hasher func(data ...[]byte) []byte, defaultHashes [][]byte) []byte {
+	var h []byte
+	if (n.left() == nil || len(n.left().val) == 0) && (n.right() == nil || len(n.right().val) == 0) {
+		//s.deleteOldNode(oldRoot) //TODO
```
howardjohn, line 247: what is the impact of this todo?

howardjohn (approved):
I think I found my next interview question :slightly_smiling_face:

howardjohn:
Put a hold since you said you hadn't tested in istiod yet

ledger/ledger.go
therealmitchconnors: This was useful previously when we were setting expiration for non-current nodes, but the Erase() function makes that unnecessary.  Honestly, we don't need this to be an expiring cache at all now, but this library doesn't have a simple cache, and when it was written Martin was requiring that all cache usages in istio use this library.  Not sure why.

I can move the evictionInterval to a day if that helps...

ledger/ledger.go
therealmitchconnors: It is used in newSMT, only because we don't have a simple cache to use...

ledger/node.go
therealmitchconnors: Ah, my IDE collapsed this line, and I I couldn't see the comment.   This was a reminder to make sure old nodes get expired.  It is handled by Erase().

therealmitchconnors:
> I think I found my next interview question 🙂

I would fail that interview so badly.

ledger/ledger.go
howardjohn: Moving to a day doesn't help, I think that would either lead to no benefits or lead to obscure issues at a day?

maybe we can just tweak it to not run the goroutine if there is no ttl set

therealmitchconnors: Oh, yeah, we can just set it to 0.  Will do.

https://github.com/istio/pkg/blob/master/cache/ttlCache.go#L99

PR #341, opened 2021-04-14
add support for scoped rate limits on logging

douglas-reid:
This PR adds a new option for `Scope`s, allowing a Scope to adopt a rate-based limit on logging messages. The option enables the ability to restrict the number of the same log lines emitted for a given period of time.

This PR is motivated by a recent customer experience with a misconfiguration resulting in an excessive amount of repetitive logs at high-frequency. The intent is to allow Istio devs to control the amount of logs they generate in control loops, etc.

I wanted to make this `Scope`-based to avoid having to adopt a single general rate across all scopes in each Istio components, and to allow incremental tightening of logs-driven noise. There are a number of possible alternatives (including per-log-line limits), but this approach seemed to me to present a decent compromise between complexity and utility. 

Example of noisy logs: 
image

A use case would be updating the `validationController` scope to only log once every 5 minutes or so.

### DIFF
--- a/log/scope.go
+++ b/log/scope.go
@@ -21,6 +21,7 @@ import (
 	"sync/atomic"
 
 	"golang.org/x/time/rate"
+
 	"istio.io/pkg/structured"
 )
 
--- a/log/scope.go
+++ b/log/scope.go
@@ -20,6 +20,7 @@ import (
 	"sync"
 	"sync/atomic"
 
+	"golang.org/x/time/rate"
 	"istio.io/pkg/structured"
 )
 
@@ -58,6 +59,12 @@ type Scope struct {
 	// labels data - key slice to preserve ordering
 	labelKeys []string
 	labels    map[string]interface{}
+
+	// sampling
+	sample bool
+	limit  rate.Limit
+	burst  int
+	limits sync.Map
 }
 
 var (
@@ -302,6 +309,9 @@ func (s *Scope) Debugf(args ...interface{}) {
 		if len(args) > 1 {
 			msg = fmt.Sprintf(msg, args[firstIdx+1:]...)
 		}
+		if !s.isSampled(msg) {
+			return
+		}
 		s.callHandlers(DebugLevel, s, ie, msg)
 	}
 }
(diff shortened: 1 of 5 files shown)


log/scope.go
```diff
@@ -387,12 +398,37 @@ func (s *Scope) WithLabels(kvlist ...interface{}) *Scope {
 	return out
 }
 
+func (s *Scope) WithRateSampling(limit rate.Limit, burst int) *Scope {
```
howardjohn, line 401: Would be good to comment on here what happens if we hit the limit - does it stall, drop the write, async write later, etc

log/scope.go
```diff
@@ -302,6 +310,9 @@ func (s *Scope) Debugf(args ...interface{}) {
 		if len(args) > 1 {
 			msg = fmt.Sprintf(msg, args[firstIdx+1:]...)
 		}
+		if !s.isSampled(msg) {
```
howardjohn, line 313: is it intentionally only at debug level?

howardjohn (commented):
If our only usage of this is for the validation logging, should we just make that not output 10 logs per second (which includes api-server calls, IIRC)? Perhaps this is generally useful as well but it feels a bit like hiding the problem rather than fixing it?

log/scope.go
douglas-reid: good catch - i forgot to remove that when I moved the callsite.

log/scope.go
douglas-reid: Added a bit of commentary. please let me know what you think.

douglas-reid:
> If our only usage of this is for the validation logging, should we just make that not output 10 logs per second (which includes api-server calls, IIRC)? Perhaps this is generally useful as well but it feels a bit like hiding the problem rather than fixing it?

I started down that approach, but then quickly wondered if this would be generally useful -- and wouldn't involve as many conversations around "do we really need this log?".  I also was a bit worried about hacking around the validation reconcile loop and wanted something a bit more "graceful".

Open to other opinions on this however.

howardjohn:
> I started down that approach, but then quickly wondered if this would be generally useful -- and wouldn't involve as many conversations around "do we really need this log?". I also was a bit worried about hacking around the validation reconcile loop and wanted something a bit more "graceful".

I think the problem is that the validation is genuinely acting too rapidly - the logs are just the visible side effect of that. There are probably other cases of places where we log too much but we actually want to do the action that is triggering the logs at that frequency otherwise though?

douglas-reid:
This never seems to have been prioritized or re-requested. Closing.

PR #358, opened 2021-05-25
Add support for int64 measures

douglas-reid:
There have been a few use cases for having the value type of metric be known as an int64 inside of the client libraries. This PR is meant to provide such a mechanism by adding an option for using int64 values, in conjunction with a new `RecordInt()` method on the interface.

If the existing `Record()` method is used on a Metric that requested int64 values, the value will be converted to int64 by taking the `math.Floor` of the value.

### DIFF
--- a/monitoring/monitoring.go
+++ b/monitoring/monitoring.go
@@ -17,6 +17,7 @@ package monitoring
 import (
 	"context"
 	"fmt"
+	"math"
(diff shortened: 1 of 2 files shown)


bianpengyuan (approved):
looks great, thanks!

howardjohn (approved):

PR #368, opened 2021-06-15
Add vendor_version for build info

shamsher31:
To fix https://github.com/istio/istio/issues/33421


mandarjog:
@shamsher31 in the build if VERSION is set it is used to link stamp binaries. We will now set that VERsIOn  to the full version including Google specific extensions. Like 1.10.2-asm.5

PR #373, opened 2021-06-18
Add log tee to a UDS server.

bianpengyuan:
Example usage https://github.com/bianpengyuan/istio/commit/259d2f4d134b76fd9386393c571b8f771a99b4a4

For https://github.com/istio/istio/issues/32437


log/uds.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// Write implements zapcore.Core. It writes a log entry to an UDS server.
+func (u *udsCore) Write(entry zapcore.Entry, fields []zapcore.Field) error {
+	attempts := 0
+	b := backoff.NewExponentialBackOff()
+	var lastError error
+	for attempts < u.maxAttempts {
+		attempts++
+		resp, err := u.client.Post(u.url, "text/plain", strings.NewReader(entry.Message))
```
howardjohn, line 90: HTTP for logs seems slightly suspicious. There is definitely some prior art here with k8s/containers - is there a simpler approach?

This is probably fine since we have low load so I wouldn't change it for something more complex though

bianpengyuan: > There is definitely some prior art here with k8s/containers - is there a simpler approach?

If you are talking about forwarding container log to storage backend, like fluentd, I think they deal with log files. However it is bit complicated with CNI plugin, because there can be multiple CNI plugin process running simultaneously and write to the same file, which is not handled by file rotation library like https://github.com/bdurand/lumberjack, and log file read at CNI container is also going to be a bit complicated considering file rotation. So feels to me using rpc is probably easier considering load is small.

Re the protocol, these are the protocols supported by fluent: https://docs.fluentd.org/input seems like all other protocols other than http are going to add quite some complexity, like customized codec on the log stream, or import some uncommon fluent golang library. Both are less ideal than just use plain http.

howardjohn: Ok, seems reasonable to me

log/uds.go
```diff
@@ -0,0 +1,109 @@
...
+			DialContext: func(_ context.Context, _, _ string) (net.Conn, error) {
+				return net.Dial("unix", address)
+			},
+		},
+		Timeout: 100 * time.Millisecond,
+	}
+	uc := &udsCore{client: c, url: "http://unix" + path, maxAttempts: maxRetryAttempts + 1}
+	for l := zapcore.DebugLevel; l <= zapcore.FatalLevel; l++ {
+		if baseCore.Enabled(l) {
+			uc.minimumLevel = l
```
howardjohn, line 50: isn't this per-scope?

log/uds.go
```diff
@@ -0,0 +1,109 @@
...
+
+// Sync implements zapcore.Core.
+func (u *udsCore) Sync() error {
+	return nil
+}
+
+// Write implements zapcore.Core. It writes a log entry to an UDS server.
+func (u *udsCore) Write(entry zapcore.Entry, fields []zapcore.Field) error {
+	attempts := 0
+	b := backoff.NewExponentialBackOff()
```
howardjohn, line 86: nit: Do we need backoff? Seems like it might be excessive given its over local UDS, and we can be lossy(?)

log/uds.go
```diff
@@ -0,0 +1,109 @@
...
+	return ce
+}
+
+// Sync implements zapcore.Core.
+func (u *udsCore) Sync() error {
+	return nil
+}
+
+// Write implements zapcore.Core. It writes a log entry to an UDS server.
+func (u *udsCore) Write(entry zapcore.Entry, fields []zapcore.Field) error {
```
howardjohn, line 84: I am guessing Write() is call synchronously? Any concerns of calls to log going from printing to stdout (Fast) to making an HTTP call (slow, especially with retry)? 

it seems like we may want async

bianpengyuan: Yeah I considered this. CNI process is short lived, and we might not even get time to finish a report cycle. Let me see if there is a way to make a final report call at process exiting at log package.

log/uds.go
douglas-reid: is UDP not simpler and lower overhead?  WriteToUDP / ReadFromUDP()?

bianpengyuan: > is UDP not simpler and lower overhead? WriteToUDP / ReadFromUDP()?

I'd rather not invent a new protocol for the log stream in order to use UDP. IMO HTTP performance is OK considering the low load.

douglas-reid: Do we need a protocol beyond write msg bytes and read msg bytes?

Should we add a bunch of documentation around the load, etc., to the option if we do use HTTP POSTs, so that non-CNI users understand the intent?

bianpengyuan: > Do we need a protocol beyond write msg bytes and read msg bytes?

Yeah probably not, we could put everything into a JSON obj and pass it as a message. However, I am updating this to buffer & send, with that, connection overhead should not be considerable? HTTP vs something + UDP probably does not matter any more.

douglas-reid: OK.

douglas-reid:
What else do we need to do to help move this along?

log/uds.go
bianpengyuan: Scope is one level above. The minimumLevel here is always debug level if I understand correctly.

log/uds.go
bianpengyuan: Removed.

log/uds.go
bianpengyuan: Make it async.

bianpengyuan:
> What else do we need to do to help move this along?

@douglas-reid Made several updates. please take another look, thanks!

log/config.go
```diff
@@ -345,6 +347,20 @@ func Configure(options *Options) error {
 		}
 	}
 
+	if options.teeToUDSServer {
+		// build uds core.
+		core = teeToUDSServer(core, options.udsSocketAddress,
+			options.udsServerPath)
```
douglas-reid, line 353: nit: do we need to wrap these lines (here and below)? it is hard to tell in github, but if not a linter violation, let's make this one line.

log/options.go
```diff
@@ -159,6 +164,13 @@ func (o *Options) WithTeeToStackdriver(project, logName string, mr *monitoredres
 	return o
 }
 
+func (o *Options) WithTeeToUDS(addr, path string) *Options {
```
douglas-reid, line 167: nit: can we add a comment here about the protocol and the use cases, so that anyone that stumbles on this option can understand more of what this is about?

douglas-reid (commented):
Looks pretty good.

log/config.go
bianpengyuan: Done.

log/options.go
bianpengyuan: Done.

douglas-reid (approved):

bianpengyuan:
@howardjohn please take a look.

howardjohn (approved):

log/config.go
```diff
@@ -345,6 +347,18 @@ func Configure(options *Options) error {
 		}
 	}
 
+	if options.teeToUDSServer {
+		// build uds core.
+		core = teeToUDSServer(core, options.udsSocketAddress, options.udsServerPath)
+		if err != nil {
```
believening, line 353: returns without error but check it

PR #374, opened 2021-06-18
Export NewLogrAdapter

therealmitchconnors:
I'd like to use this for building controllers which use logr, but not klog.

### DIFF
--- a/log/config.go
+++ b/log/config.go
@@ -389,7 +389,7 @@ func Configure(options *Options) error {
 
 	// capture klog (Kubernetes logging) through our logging
 	configureKlog.Do(func() {
-		klog.SetLogger(newLogrAdapter(KlogScope))
+		klog.SetLogger(NewLogrAdapter(KlogScope))
 	})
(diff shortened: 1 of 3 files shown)


douglas-reid:
I don't see a big issue with this PR. That being said, what are the tradeofss vs. having a Config option to opt-out of klog ?

therealmitchconnors:
> I don't see a big issue with this PR. That being said, what are the tradeofss vs. having a Config option to opt-out of klog ?

I'm not sure.  To be clear, this function is only currently used for configuring klog, and this PR allows us to use it for things other than klog.  It has no affect on the current klog use cases.

howardjohn:
SGTM if Doug has no concerns

howardjohn (approved):
@douglas-reid Feel free to remove hold once your concerns are addressed

mandarjog (approved):
lgtm. Just exposing existing methods.

douglas-reid (approved):

PR #384, opened 2021-07-10
Add lock on operations to UDS log buffer.

bianpengyuan:
Unit test in https://github.com/istio/istio/pull/33926 detects race condition. Add lock on any buffer operations.

### DIFF
--- a/log/uds.go
+++ b/log/uds.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"sync"
 	"time"
 
 	"go.uber.org/zap/buffer"
@@ -34,6 +35,7 @@ type udsCore struct {
 	url          string
 	enc          zapcore.Encoder
 	buffers      []*buffer.Buffer
+	mu           sync.Mutex
 }
 
 // teeToUDSServer returns a zapcore.Core that writes entries to both the provided core and to an uds server.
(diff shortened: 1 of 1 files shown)


log/uds.go
```diff
@@ -111,6 +115,8 @@ func (u *udsCore) Write(entry zapcore.Entry, fields []zapcore.Field) error {
 	if err != nil {
 		return fmt.Errorf("failed to write log to uds logger: %v", err)
 	}
+	u.mu.Lock()
+	defer u.mu.Unlock()
```
mandarjog, line 119: Not: no advantage in using defer instead of unlocking before return.

log/uds.go
```diff
@@ -84,12 +86,14 @@ func (u *udsCore) Check(e zapcore.Entry, ce *zapcore.CheckedEntry) *zapcore.Chec
 
 // Sync implements zapcore.Core. It sends log messages with HTTP POST.
 func (u *udsCore) Sync() error {
+	u.mu.Lock()
```
mandarjog, line 89: Nit: consider a private func “logsFromBuffer()” which will do locking in the beginning and unlocking at the end. Since it has functions that can panic defer may be appropriate in that function.

mandarjog (approved):
Lgtm

log/uds.go
bianpengyuan: Done.

log/uds.go
bianpengyuan: Done.

PR #401, opened 2021-07-26
Add support for Close to logging package

douglas-reid:
When `TeeToStackdriver` type logging is requested, clients can be established to handle the buffering and pushing of logs to the Cloud Logging service. However, there is no existing way to close these clients. This can lead to leaks, as detected by leak tests.

This PR seeks to add a way to ensure those clients can be closed appropriately.

### DIFF
--- a/log/config.go
+++ b/log/config.go
@@ -101,6 +101,7 @@ type patchTable struct {
 	sync        func() error
 	exitProcess func(code int)
 	errorSink   zapcore.WriteSyncer
+	close       func() error
 }
 
 var (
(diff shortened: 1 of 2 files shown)


douglas-reid:
@howardjohn @bianpengyuan wdyt about something like this to handle `Close`ing the SD clients?

howardjohn (commented):
Seems good to me

bianpengyuan (approved):
lgtm

howardjohn (approved):

PR #416, opened 2021-09-14
Optimize open-census usage

howardjohn:
See https://github.com/census-instrumentation/opencensus-go/issues/1265

Benchmark:
```
name       old time/op    new time/op    delta
Counter-6     713ns ± 8%     680ns ± 3%   -4.63%  (p=0.006 n=10+9)

name       old alloc/op   new alloc/op   delta
Counter-6      205B ± 0%      173B ± 0%  -15.63%  (p=0.000 n=10+10)

name       old allocs/op  new allocs/op  delta
Counter-6      6.00 ± 0%      5.00 ± 0%  -16.67%  (p=0.000 n=10+10)
```

This is not *that* much faster, but once my OC PRs merge we can take advantage of more optimizations this way, bringing us down to:
```
name       old time/op    new time/op    delta
Counter-6     713ns ± 8%     618ns ± 6%  -13.35%  (p=0.000 n=10+9)

name       old alloc/op   new alloc/op   delta
Counter-6      205B ± 0%       68B ± 1%  -66.59%  (p=0.000 n=10+10)

name       old allocs/op  new allocs/op  delta
Counter-6      6.00 ± 0%      3.00 ± 0%  -50.00%  (p=0.000 n=10+10)
``

### DIFF
--- a/monitoring/monitoring.go
+++ b/monitoring/monitoring.go
@@ -231,8 +231,14 @@ func (d *derivedFloat64Metric) Register() error {
 type float64Metric struct {
 	*stats.Float64Measure
 
+	// tags stores all tags for the metrics
 	tags []tag.Mutator
+	// ctx is a precomputed context holding tags, as an optimization
+	ctx  context.Context
 	view *view.View
+
+	incrementMeasure []stats.Measurement
+	decrementMeasure []stats.Measurement
 }
 
 func createOptions(opts ...Options) *options {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: monitoring/monitoring.go [hunk 1]
```diff
@@ -231,8 +231,14 @@ func (d *derivedFloat64Metric) Register() error {
 type float64Metric struct {
 	*stats.Float64Measure
 
+	// tags stores all tags for the metrics
 	tags []tag.Mutator
+	// ctx is a precomputed context holding tags, as an optimization
+	ctx  context.Context
```
douglas-reid, line 237: This technically violates the explicit documentation for `context.Context`:

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it.

But I'm not sure that matters for this use case. This is a vestige of the OC library construction, and is not holding request-scoped information or any sort of deadlines, etc.

douglas-reid (approved):
Looks OK to me. Thanks for tackling!

ericvn (approved):

PR #446, opened 2021-12-14
Fix use of klog/v2 to use logr.Logger struct instead of interface

jacob-delgado:


jacob-delgado:
I still need to do some testing to see that this works as intended. How do I test this @howardjohn / @ericvn ?

howardjohn:
you can pull it into Istio/istio and then add --vklog 6 to istioctl and it
should log in our format. We do have pretty decent test coverage here as
well

log/logr.go
```diff
@@ -76,27 +77,27 @@ func (zl *zapLogger) Error(err error, msg string, keysAndVals ...interface{}) {
 	}
 }
 
-func (zl *zapLogger) V(level int) logr.Logger {
-	return &zapLogger{
-		lvl:    zl.lvl + level,
-		l:      zl.l,
-		lvlSet: true,
+func (zl *zapLogger) V(int) logr.Logger {
```
howardjohn, line 80: I am a bit suspicious that we don't use the level here? How does that work?

howardjohn (approved):
Seems to work but I don't actually understand how the V() stuff works :-)

PR #455, opened 2022-01-17
use local time instead of utc

wccsama:


howardjohn (changes requested):
https://github.com/istio/istio/issues/30550

PR #490, opened 2022-03-01
Add basic label support for derived metrics

douglas-reid:
This is a proposal for a mechanism to support derived gauges that take label dimensions. This would allow the creation of a derived gauge that can be dimensioned with differing derivations of value. For instance, cert expiry metrics could be created with differing values based on the cert name or trust domain, etc., but with the same overall metric name.

I'm looking for feedback on overall approach and naming.

### DIFF
--- a/monitoring/monitoring.go
+++ b/monitoring/monitoring.go
@@ -269,14 +269,18 @@ func (d *derivedFloat64Metric) Register() error {
 
 func (d *derivedFloat64Metric) ValueFrom(valueFn func() float64, labelValues ...string) {
 	if len(labelValues) == 0 {
-		d.base.UpsertEntry(valueFn)
+		if err := d.base.UpsertEntry(valueFn); err != nil {
+			log.Errorf("failed to add value for derived metric %q: %v", d.name, err)
+		}
 		return
 	}
 	lv := make([]metricdata.LabelValue, 0, len(labelValues))
 	for _, l := range labelValues {
 		lv = append(lv, metricdata.NewLabelValue(l))
 	}
-	d.base.UpsertEntry(valueFn, lv...)
+	if err := d.base.UpsertEntry(valueFn, lv...); err != nil {
+		log.Errorf("failed to add value for derived metric %q: %v", d.name, err)
+	}
 }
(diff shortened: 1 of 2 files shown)


monitoring/monitoring.go
```diff
@@ -186,28 +219,23 @@ func NewGauge(name, description string, opts ...Options) Metric {
...
-	return NewDerivedGaugeWithLabels(name, description, nil, valueFn)
-}
-
-// NewDerivedGaugeWithLabels creates a new Metric that takes its value from the provided function with a set of const labels.
-func NewDerivedGaugeWithLabels(name, description string, labels map[string]string, valueFn func() float64) DerivedMetric {
-	constLabels := make(map[metricdata.LabelKey]metricdata.LabelValue)
-	for k, v := range labels {
-		constLabels[metricdata.LabelKey{Key: k}] = metricdata.NewLabelValue(v)
-	}
+func NewDerivedGauge(name, description string, opts ...DerivedOptions) DerivedMetric {
```
zirain, line 222: Overall, this work.
```console
istio_agent_cert_expiry_seconds{resource_name="default"} 86354.308602142
istio_agent_cert_expiry_seconds{resource_name="default1"} 86354.308604592
```

BUT, this changes will broken all old implements.
If there's an automator, that will failed on here 
https://github.com/istio/istio/blob/b020516b7ddcfc19f4b635a3c85a8eb7b25b2b12/pilot/pkg/bootstrap/monitoring.go#L44

howardjohn: That is ok, we do not need backwards compatibility in this repo

douglas-reid: It will require a manual update of the dep for istio/istio. I don't think the `DerivedGauge` stuff has been used anywhere other than that single location, so it should be easy to update hopefully.

zirain: I will update istio/istio once this is merged.

zirain:
@howardjohn can we get this merged?

howardjohn:
I was waiting for you to approve. I assume it looks good to you?

PR #605, opened 2022-08-04
Enable configuring klog without cobra

howardjohn:
Useful for integration tests, etc which don't use cobra


ericvn (approved):

ericvn:
Looks like you need to change a few ioutil.* to os.*

PR #620, opened 2022-08-18
add support for conditional registration of metrics

ramaraochavali:


monitoring/monitoring.go
```diff
@@ -63,6 +63,10 @@ type (
 		// Register configures the Metric for export. It MUST be called before collection
 		// of values for the Metric. An error will be returned if registration fails.
 		Register() error
+
+		// disable configures the metric in 'Disabled' state. It is useful for cases where
+		// metrics are enabled based on some configuration flag.
+		disable()
```
howardjohn, line 69: TIL interfaces can have private methods. Makes sense, I had just never seen it before.

monitoring/monitoring.go
```diff
@@ -460,6 +515,10 @@ func (i *int64Metric) RecordInt(value int64) {
 	stats.Record(i.ctx, i.M(value)) //nolint:errcheck
 }
 
+func (i *int64Metric) disable() {
+	i.disabled = true
+}
+
 func (i *int64Metric) With(labelValues ...LabelValue) Metric {
 	t := make([]tag.Mutator, len(i.tags), len(i.tags)+len(labelValues))
```
howardjohn, line 523: nit: do we want to make this a NOP for disabled? `with` is pretty slow so that make improve performance?

monitoring/monitoring_test.go
```diff
@@ -69,10 +69,24 @@ var (
...
+	testConditionalSum = monitoring.NewSum(
+		"events_conditional_total",
+		"Number of events observed, by name and kind",
+		monitoring.WithLabels(name, kind),
+	)
 )
 
 func init() {
 	monitoring.MustRegister(testSum, hookSum, int64Sum, testDistribution, testGauge)
+	monitoring.RegisterIf(testDisabledSum, func() bool { return false })
```
howardjohn, line 88: I think this works as well - definitely fine merging. One thing I was also considering is if we wanted to make it conditional in `NewSum` instead of register? Then we could have a `disabledMetric struct` instead of `if disabled { ... }` everywhere.

howardjohn (approved):
I put a hold since I had comments but I don't mind merging

monitoring/monitoring.go
ramaraochavali: Makes sense. Will change

monitoring/monitoring_test.go
ramaraochavali: I was thinking about that but was in two minds - but if you prefer, I can change it to NewSum

howardjohn: I like the usage (from ux) of RegisterIf but the implementation of NewSum :slightly_smiling_face:. I think this is probably fine

ramaraochavali: Changed RegisterIf to return the registered metric or disabledMetric so that we can use disabledMetric. See if this better.

ramaraochavali:
@howardjohn can you PTAL? Are we good to remove the hold?

howardjohn:
I didn't look recently but feel free to remove it. Or I'll look Monday and
remove

ramaraochavali:
Not urgent. Please do it on Monday

monitoring/monitoring_test.go
```diff
@@ -69,10 +69,24 @@ var (
...
 		"Testing gauge functionality",
 	)
+
+	testDisabledSum = monitoring.NewSum(
+		"events_disabled_total",
+		"Number of events observed, by name and kind",
+		monitoring.WithLabels(name, kind),
+	)
+
+	testConditionalSum = monitoring.NewSum(
```
howardjohn, line 79: curious why we register and init seperately. Could we just do `testConditionalSum = monitoring.Register(monitoring.NewSum)`? And RegisterIf where appropriate, of course?

howardjohn (approved):

monitoring/monitoring_test.go
ramaraochavali: I think we can do. Not sure what the history was

PR #644, opened 2022-10-08
Update to go 1.19 in go.mod

hanxiaop:


howardjohn (commented):
Do we need this? This really only makes it so you cannot build proper on 1.18 afaik

ericvn:
I agree with @howardjohn and not sure why this is needed. Certainly we aren't adding 1.19 function in this PR. Although I'm not sure who might still be using 1.18 with istio/pkg, this will prevent that from working.

PR #659, opened 2022-10-21
Update dependencies

jacob-delgado:


go.mod
```diff
@@ -1,57 +1,57 @@
 module istio.io/pkg
 
-go 1.18
+go 1.19
```
ericvn, line 3: We had an earlier discussion about if we really want to update to 1.19 here. This requires anyone using this to also be using 1.19 for their builds.

jacob-delgado: istio/istio uses go 1.19. I'll just change it to 1.18. There's no requirement for us to update.

howardjohn (approved):

PR #663, opened 2022-10-30
chore(deps): use modules instead of incompatible

mmorel-35:
This replace github.com/natefinch/lumberjack incompatible version with gopkg.in/natefinch/lumberjack.v2 module and 
gotest.tools incompatible version with gotest.tools/v3 module 

Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>


howardjohn (commented):
this seems reasonable but I am curious what motivated the change?

mmorel-35:
Incompatibles libraries aren't go modules and as istio/pkg is a module it seems more sustainable to depend on modules when they are availables. IMHO it's seems like using modules a cleaner approach, as you can see it allows to remove some indirect dependencies.

howardjohn (approved):
/ok-to-test

PR #787, opened 2023-04-18
Add proxy type to proxy info

hanxiaop:
This is useful in some places to diffrenciate sidecar and ztunnel proxies.

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -58,8 +58,8 @@ const (
 	// SidecarProxy type is used for sidecar proxies in the application containers
 	SidecarProxy NodeType = "sidecar"
 
-	// Router type is used for standalone proxies acting as L7/L4 routers
-	Router NodeType = "router"
+	// Gateway type is used for gateway proxies that enable ingress and egress traffic
+	Gateway NodeType = "gateway"
 
 	// Waypoint type is used for waypoint proxies
 	Waypoint NodeType = "waypoint"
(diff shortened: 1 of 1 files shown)


version/version.go
```diff
@@ -55,6 +55,7 @@ type MeshInfo []ServerInfo
 type ProxyInfo struct {
 	ID           string
 	IstioVersion string
+	Type         string
```
howardjohn, line 58: can we define a few types as enum?

hanxiaop: Just copied & pasted the `NodeType` from the istio repo.

version/version.go
```diff
@@ -51,10 +51,28 @@ type ServerInfo struct {
...
 
+// NodeType decides the responsibility of the proxy serves in the mesh
+type NodeType string
+
+const (
+	// SidecarProxy type is used for sidecar proxies in the application containers
+	SidecarProxy NodeType = "sidecar"
+
+	// Router type is used for standalone proxies acting as L7/L4 routers
+	Router NodeType = "router"
```
howardjohn, line 62: can we call it gateway? users do not really know what router is, its more of a legacy term for internals.

howardjohn (approved):

howardjohn:
but dnm for small comment

version/version.go
hanxiaop: Thanks, revised.

hanxiaop:
The comment is resolved, and I'll remove dnm

PR #795, opened 2023-05-01
log: simplify API

howardjohn:
Remove callerSkip from RegisterScope; it should always be 0.

Remove ability to pass multiple values to non `f` types (log.Info). This removes a number of bugs in usage, as calls need to be very careful about spacing. Most callers should use structured logs.

This simplicity will make moving to
https://github.com/istio/istio/issues/44683 easier as well.

Istio PR: https://github.com/istio/istio/pull/44684

### DIFF
--- a/log/config.go
+++ b/log/config.go
@@ -300,7 +300,7 @@ func processLevels(allScopes map[string]*Scope, arg string, setter func(*Scope, 
 			}
 			return nil
 		} else if s == GrpcScopeName {
-			grpcScope := RegisterScope(GrpcScopeName, "", 3)
+			grpcScope := registerScope(GrpcScopeName, "", 3)
 			logGrpc = true
 			setter(grpcScope, l)
 			return nil
--- a/log/config_test.go
+++ b/log/config_test.go
@@ -91,7 +91,7 @@ func TestTimestampProperMicros(t *testing.T) {
 
 func TestOverrides(t *testing.T) {
 	resetGlobals()
-	s := RegisterScope("TestOverrides", "For testing", 0)
+	s := RegisterScope("TestOverrides", "For testing")
 
 	o := DefaultOptions()
 	o.outputLevels = "default:debug,all:info"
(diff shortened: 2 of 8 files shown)


costinm (commented):
I'm not sure I understand the purpose of this.

If log.Warn/Error/etc is commonly used with a single param - because the extra params had the space issue - why not just fix it to behave like golang log and add the space ?

costinm:
Also not sure what process we follow for the pkg repo - but generally backward incompatible changes to an API are not recommended. If this is internal istio impl that should not be used standalone - probably should not be presented as a separate pkg with independent versioning. 

I would revert the part where the ...any is removed - and instead fixed it to add the spaces and behave closer to 
upstream (and other) logging APIs, or use it for structured logging.

howardjohn:
Changing to add space will break everyone, including istio/istio, which
simultaneously depends on it using spaces AND not using spaces (we have
some invalid callings).

istio/pkg is not intended to be used outside of Istio and I would be in
favor of removing it entirely.

costinm:
+1 on removing it as soon as slog is out.

howardjohn:
> I would revert the part where the ...any is removed - and instead fixed it to add the spaces and behave closer to


Just to clarify why I did this - it was to be future compat with slog.

If we move `Info(...any)` from "concat args" to "structured logs", we will break logs.

But if you move `Info(any)` to `Info(...any)` with structured logging, it will not break anyone, since no once could have multiple arguments. So only new usage would hit the new structured logging code.

costinm:
There are only a couple of methods with log.Info(...), most use Infof.
The code in the log package is pretty bad in terms of allocs, copies and in
general - there is a good part in
registering the logs and dynamically setting levels which we should
preserve. And the Info() adding params
without spaces is the strangest thing, I don't disagree that we should stop
using it - just how.

I would first change istio/ repo to replace the usages - but as we do this
we may as well make the
log statements structured.

How about adding InfoCtx ( and similar ) - so it is super clear and we
start passing the context.
New code and as we update can use this method, we gradually move - and then
kill Info(...any) ?

PR #796, opened 2023-05-01
cleanup: move bindata gen to embed

howardjohn:


jacob-delgado (approved):
can we remove go-bindata from istio/tools?

howardjohn:
Some other repos depend on it. But we can move all of them off as well,
just needs some grunge work

PR #804, opened 2023-05-08
Separate OpenCensus and remove unused methods

costinm:
This is still WIP (commented out a number of unused methods, will cleanup). 

Like the log change, there are a few small updates in istio repo, but very small. Almost all of Istio is using
the float variables and the methods removed are not used and will simplify migration from OC.


monitoring/monitoring.go
```diff
@@ -43,25 +37,42 @@ type (
...
+		//
+		// Not used in istio, removed pending new interface.
+		//Decrement()
 
 		// Name returns the name value of a Metric.
+		// TODO: internal use only
 		Name() string
 
 		// Record makes an observation of the provided value for the given measure.
+		// Majority of Istio is setting this to an int
```
howardjohn, line 49: This doesn't imply we are going to change it to `int`, right? Since all the histograms are floats (duration)

monitoring/monitoring.go
```diff
@@ -43,25 +37,42 @@ type (
...
+		// Not actually used in Istio.
+		//RecordInt(value int64)
+
+		// Prometheus uses ConterVec and With(map[string]string) returning Counter
+		//
+		// Expvar doesn't support labels - but can be used in the name, creating a new expvar in a map
+		// The labels are really made part of the name when exporting and is the name of the TS
+		//
+		// Otel uses attribute.Int("name", val) and similar - from the stable package -
+		// but doesn't create a new Metric, it is an option to Add().
```
howardjohn, line 62: Not pre-creating objects with the attributes is what was costing Istiod to spend ~25% of its time in metrics code. So hopefully this is not suggesting we do that?

monitoring/monitoring.go
```diff
@@ -254,95 +264,69 @@ func RegisterIf(metric Metric, enabled func() bool) Metric {
...
-	return newMetric(name, description, view.Sum(), opts...)
+	return newSum(name, description, opts...)
 }
 
 // NewGauge creates a new Metric with an aggregation type of LastValue. That means that data collected
 // by the new Metric will export only the last recorded value.
 func NewGauge(name, description string, opts ...Options) Metric {
-	return newMetric(name, description, view.LastValue(), opts...)
+	if strings.HasSuffix(name, "_total") {
+		fmt.Println(name)
```
howardjohn, line 306: Should rm before merge

monitoring/monitoring_opencensus.go
```diff
@@ -0,0 +1,229 @@
+//go:build !skip_opencensus
```
howardjohn, line 1: Whats the long term plan for this? its not easy to set build tags in istio release today. But not a big deal if its short lived

monitoring/monitoring.go
costinm: No, I'm removing ints from this API since Istio is only using floats. Just noticing it...

monitoring/monitoring_opencensus.go
costinm: Since OC is deprecated - it has to be short lived :-)

A separate PR will add an implementation based on prom client, and swap the default - I'm not sure if we should have an env variable to control or just switch, will get there - one step at a time, but I do not plan to have this in the release.

howardjohn: Why prom? I thought we were moving to otel?

costinm:
OTel metrics are not yet 'stable' - while prom is a dependency we already
have. Even with otel API, exporting to prometheus is still based on the
prometheus client library, by registering a 'generator'.

We can still use otel traces which are v1+ - but in practice registering a
counter/gauge/historgram with the prometheus client library will continue
to work even if we start
registering some counters with OTel API when it moves to v1 - and we have
an opportunity to validate the performance and impact.

We can also start using OTel collector - which supports reading prom
metrics and pushing to OTLP endpoints.

monitoring/monitoring_opencensus.go
kyessenov: Why not use otel SDK to allow exporters besides prometheus? Semantically they should be equivalent per their stated goal.

costinm:
Few reasons:
1. Otel metric SDK is not yet 1.0/stable
2. "Stated goal" may or may not be equivalent with reality - we have a lot
of stated goals too
3. It is an interface - best to have multiple implementations, so we can
actually compare for ourselves.
4. Not clear if Istio should take the burden of configuring other exporters
(and their auth) instead of letting OTel-collector
or similar tools handle that - across k8s including non-istio workloads.
Same applies to istiod.

When Otel metric is stable/1.0 and we check the numbers - we may as well
start using the OTel API directly
instead of a wrapper ( I think we should do the same for log once new slog
is available ).

kyessenov:
SGTM, as long as we maintain a wrapper and don't force any library until Otel is ready.

costinm:
Agreed, just like we want Istio users to only use stable features in
production and not experimental - we should do the same process for our
dependencies :-)

And even when OTel metrics is v1 - we would still want to test and compare
it with the prom library which is in broad use. Whatever is faster
and better can be used ( unless prometheus library gets deprecated ).

OTel tracing is v1 - and we can use it, and I believe for access logs there
is some gray area. There is also a subtle difference between
protocol and the client libraries - supporting OTLP and OTel is possible
without using a particular library.

PR #820, opened 2023-06-05
fix ledger delete func

shirleyyLi:
To fix https://github.com/istio/istio/issues/41655


jacob-delgado:
This repo is deprecated. While it's fine to fix it here, can you also put up a PR for this change in istio/istio at https://github.com/istio/istio/blob/bcd3fb233aea77ee6f32bbf8043370449175d330/pkg/ledger/ledger.go#L64?

shirleyyLi:
@jacob-delgado Thanks. I've created a new pr https://github.com/istio/istio/pull/45346

